PR #152, opened 2014-12-13
Changed Media Queries from Pixels to ems

dbalders:
No Issue

* Changed from 900px and 500px to 56.25em and 31.25em. 
* The ems still break on the same pixels on my mac using chrome
* Figured might be worth it in case people have different base font sizes or whatever

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -1334,7 +1334,7 @@ body:not(.post-template) .post-title {
    10. Media Queries - Smaller than 900px
    ========================================================================== */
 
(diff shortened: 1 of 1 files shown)


mankittens:
With `.post` width being defined as % and px, I can't see why ems should be used for media queries.

If the width of the site were defined by em, I could see the argument, though.

dbalders:
Fair point. Just kinda stumbled across it one day and noticed that most of what was used was em, so figured I would make it match, but obviously not a big deal either way :smile:

PR #158, opened 2015-01-12
fix DISQUS iframe margin problem

Rezaaa:
on post template, DISQUS iframe have some margin and it's out of screen.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -423,7 +423,7 @@ iframe, .fluid-width-video-wrapper {
 }
 
 /* fix DISQUS iframe margin problem */
-.post-template > iframe#dsq-5 {
+iframe#dsq-5 {
    margin: 0 !important;
 }
 
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -422,6 +422,11 @@ iframe, .fluid-width-video-wrapper {
     margin: 1.75em 0;
 }
 
+/* fix DISQUS iframe margin problem */
+.post-template > iframe {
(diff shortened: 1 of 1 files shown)


JohnONolan:
Thanks for getting involved! Please could you provide a before/after screenshot of this problem and what the PR fixes?

Rezaaa:
@JohnONolan 
screenshot

JohnONolan:
The bug is valid - but that's not the right fix. That's overriding the iframe default margin on all iframes, which is not what we want. We'll need to find a way to target the disqus iframe specifically, or modify the default iframe style to be applied only to content and not elsewhere.

Rezaaa:
@JohnONolan check it again please.

JohnONolan:
This is fine - but it needs to be squashed down to 1 commit and the code needs to live in the "3rd party" section - same place where we have the custom styles for Github Gists

mankittens:
This fix should specifically target the #dsq-5 iframe. It's not an issue for all disqus iframes, as the current commit reflects.

```css
/* Disqus */
iframe#dsq-5 {
    margin: 0;
}
```

This should be sufficient, plus or minus `iframe`.

mankittens:
The `.post-template >` selector is a bit overqualified, imo.

Pherserk:
dsq-5 id could change, depending on third parties.
In my opinion is better avoiding any style directive that addresses directly the iframe tag.

JohnONolan:
@Rezaaa feel free to re-open if you have time to update

Issue #154, opened 2015-01-03
Errors in console


chenxsan:
When view http://demo.ghost.io/ with firefox console opened, I got some error messages:

![ghost](https://cloud.githubusercontent.com/assets/1091472/5602377/f7cad060-9383-11e4-9d03-291c89c6754e.jpg)

> downloadable font: OS/2: bad linegap: -32 (font-family: "casper-icons" style:normal weight:normal stretch:normal src index:1)
source: http://demo.ghost.io/assets/fonts/casper-icons.woff screen.css
> downloadable font: OS/2: bad weight: 250 (font-family: "Merriweather" style:normal weight:300 stretch:normal src index:2)
source: https://fonts.gstatic.com/s/merriweather/v8/ZvcMqxEwPfh2qDWBPxn6nmFp2sMiApZm5Dx7NpSTOZk.woff

What's the problem? Any suggest?

JohnONolan:
Hi there - I'm not able to reproduce this I'm afraid! Could you have a read of our contributing guidelines and follow the bug report template, with a particular emphasis on technical details? https://github.com/TryGhost/Ghost/blob/master/CONTRIBUTING.md#bugs

chenxsan:
Hi there, some technical details here:

+ OS - openSUSE 13.2 32 bits.
+ Browser - Firefox 34.0.5, with a lots of addons

I also tried Firefox 36.0a2 with all addons disabled, same errors in the console.

No errors in Chrome(Version 40.0.2214.45 beta) console.

JohnONolan:
Sounds like it's something to do with your OS/Browser - this isn't something we've seen before. If you find a solution, feel free to PR!

crepererum:
I can reproduce this. You have to enable CSS warning/errors to see this. It seems that these errors are mostly because of
 - outdated prefixes (e.g. `white-space: -moz-pre-wrap` which [isn't required as of Firefox 3.0](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space))
 - compatibility reasons (e.g. `background: -webkit-gradient`)
 - unsupported features, at least without prefix (e.g. `word-break: break-word;`)

But there is also 1 real error:
`.clearfix { *zoom: 1; }` at `screen.css:443` should not have this `*`, right?

JohnONolan:
Wow nice find - happy to merge if you want to pr fixes for those

crepererum:
I'll prepare one :)

JohnONolan:
[...]

PR #159, opened 2015-01-15
Fix multiple CSS problems, mostly reported by Firefox

crepererum:
- typo: `*zoom: ...;` should be `zoom: ...;`
 - outdated prefix: `white-space: -moz-pre-wrap;` is [not required as of Firefox 3.0](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space)
 - outdated prefix: `-moz-border-radius: ...;` is [not required as of Firefox 4.0](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius)
 - outdated prefix: `-webkit-border-radius: ...;` is [not required as of Chrome 4.0 and Safari 5.0](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius)
 - outdated prefix: for `background` `-moz-radial-gradient`, `-webkit-radial-gradient` and `-webkit-gradient` is [not required as of Firefox 16, Chrome 26 and Safari 6.1](http://www.w3schools.com/Css/css3_gradients.asp)
 - wrong attribute: `word-break: break-word;` should be `word-wrap: break-word;` (this makes sense because `hyphens: auto;` is set)
 - missing prefix: add `-moz-hyphens: auto;`, `-ms-hyphens: auto;` and `-webkit-hyphens: auto;` to `hyphens: auto;` (other browser [do not support `auto` mode](http://caniuse.com/#feat=css-hyphens))

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -351,7 +351,6 @@ pre {
 
 pre code, tt {
     font-size: inherit;
-    white-space: -moz-pre-wrap;
     white-space: pre-wrap;
     background: transparent;
     border: none;
@@ -440,7 +439,7 @@ margin on the iframe, cause it breaks stuff. */
     display: table;
 }
 .clearfix:after { clear: both; }
-.clearfix { *zoom: 1; }
+.clearfix { zoom: 1; }
 
 /* Hides shit */
 .hidden {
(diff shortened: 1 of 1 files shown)


crepererum:
Forgot to mention that this fixes #154

JohnONolan:
Fantastic, thank you!

PR #160, opened 2015-01-15
RFC: Add lang attribute to html element

crepererum:
## Description
This enables hyphens and text2speech. Without this responsible mobile layout is only partial activated. Especially on small screens, hyphenation can improve readability. Also text2speech is only supported by most software when the language of the content is know. So setting the `lang` attribute also improves accessibility.

Using `en` as a default language makes sense because the template is English as well. Users of other languages might want to change this.

## TODO (later)
Change hardcoded language to templated attribute when i18n is supported. See TryGhost/Ghost#3801 for more details.

## Before
lang_en_before

## After
lang_en_after

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 <head>
     {{! Document Settings }}
     <meta charset="utf-8" />


JohnONolan:
Hey there, thanks for getting involved! Always fun to see new ideas and suggestions. 

We're not going to be adding hyphens to Casper. They're very poorly supported across various browsers and generally make far more of a mess than they improve readability. Your screenshot kind of demonstrates that with 2 hyphenated words in the space of a single paragraph.

Casper doesn't have language attribute because not everybody is using Casper to write in a single language :)

crepererum:
Then you might want to change the CSS code, because you added hyphens to Casper already. See https://github.com/TryGhost/Casper/blob/master/assets/css/screen.css#L721

PR #161, opened 2015-01-16
Update post.hbs

shravanthium:


PaulAdamDavis:
Hi @shravanthium,

Casper should not have Disqus in it by default, it's something Apps will handle later down the line, or be implemented by the blog owner if they want it now.

Issue #162, opened 2015-01-21
Line number of embedded gists


janoskk:
The line numbers of embedded gists are not aligned to the code.

![screen shot 2015-01-21 at 23 38 59](https://cloud.githubusercontent.com/assets/1554533/5846766/ca9bb8a6-a1c6-11e4-824c-28bbe6915c50.png)

I have an installation of Ghost 0.5.8 + Casper 1.1.5 and tested in Chrome 39 and FF 35 (both under OS X).

JohnONolan:
Ah Github must've updated their Gist CSS or something, so now the custom CSS in casper for Gists is wrong. If you have a fix - feel free to PR and I'll be happy to merge :)

janoskk:
Here we go ;-)

JohnONolan:
Fixed in #163

PR #163, opened 2015-01-22
Fix gist line numbering issue #162

janoskk:
Please consider merging.

Tested under FF 30 & Chrome 39 (Win).

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -1185,9 +1185,11 @@ body:not(.post-template) .post-title {
     margin: 0;
     font-size: 1.4rem;
 }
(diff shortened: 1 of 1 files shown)


JohnONolan:
Thanks @janoskk - would you mind including a quick before/after screenshot here for me?

janoskk:
No problem.
Before:
screen shot 2015-01-22 at 20 08 35
After:
screen shot 2015-01-22 at 20 08 51

I tested also with a 10000 line long gist and it was aligned properly.

JohnONolan:
Awesome

PR #166, opened 2015-02-05
Loop post

DapperFox:
Added NUVI theme and started editing loop.


PaulAdamDavis:
Hi @DapperFox, given that you've changed the name in `package.json`, I guess this isn't meant to be here? Closing just in case :)

DapperFox:
@PaulAdamDavis I am bad and I should feel bad. Sorry!

PR #172, opened 2015-02-19
Improvement to .post-content .full-img

mankittens:
New code will center images and allow them to extend beyond the main text column, but not force them to be as wide as the maximum allowed width.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -858,8 +858,9 @@ body:not(.post-template) .post-title {
     border: none;
 }
 
-/* Stop .full-img from creating horizontal scroll - slight hack due to
-   imperfections with browser width % calculations and rounding */
+/* Stop elements, such as img wider than the post content, from 
+   creating horizontal scroll - slight hack due to imperfections 
+   with browser width % calculations and rounding */
 .post-template .content {
     overflow: hidden;
 }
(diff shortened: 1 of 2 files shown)


mankittens:
Just realized. This needs a small fix for mobile width. Don't accept it.

JohnONolan:
Can you explain what problem you're trying to solve with this change?

mankittens:
It's minor, admittedly, but essentially, it will cap the width at 126% of .full-img images but not force .full-img images *to be* 126%.

JohnONolan:
The current implementation only forces images to 126% if they are of sufficient resolution to go to that size. If they aren't they already stay centered at their original resolution.

mankittens:
Under the current rules, any image greater than 100% and less than 126% is stretched to 126%.

novaugust:
> Under the current rules, any image greater than 100% and less than 126% is stretched to 126%.

It's true

This works well for me, I'm using it over on my blog. 
The transform + relative positioning bit is a stroke of genius, I've been wondering for weeks how I could pop images out to the fullscreen when they're wrapped in centered 700px `<p>`s

novaugust:
@mankittens did you check your changes on a smaller screen? There's a second `.full-img` rule that plays with image width on the small screen MQ

mankittens:
I did notice that after my first commit. I went ahead and removed the `margin: 0 -16%` in the `@media max-width: 500` section, since images auto-center at any width with this fix.

You know, I just realized that these changes obsolete the need for a .full-img tag, no? (This would cut out some JavaScript in index.js!) I'll double-check in a bit.

mankittens:
Was able to cut out all references to .full-img, in index.js and style.css, since the new css rules are happy to center images big and small. Working example here: https://apatchofcode.com/cant-sign-out-of-icloud-account/.

JohnONolan:
I wasn't convinced by this initially - but you've done a really good job of demonstrating why this is a good change :+1: - If you can squash and rebase, I'd be happy to merge!

novaugust:
No more js jumping! Nice work =) Just in case you aren't sure what he means by squashing, [here's a little snippet](https://github.com/TryGhost/Ghost/wiki/Git-workflow#clean-up-history) from the Ghost contributing guide.

And i have to leave this somewhere: these changes + setting max-width to 100vw is fun.

mankittens:
Got it! Thanks for pointing me to the Ghost docs.

mankittens:
This is ready to merge!

mankittens:
Committed with smartresize and debounce removed.

JohnONolan:
Great work :+1:

PR #174, opened 2015-02-24
fixed ie9 rotate issue with scroll-down arrow

nicdutil:
In ie9, the scroll down arrow was pointing left.  I added the "-ms-transform: rotate(-90deg)" to fix the issue.

I tested it using a vmbox running win7 ie9.


JohnONolan:
Thanks :)

PR #175, opened 2015-02-24
Navigation

### LINKED ISSUE #173: Add {{navigation}} support to Casper
Initial navigation support for Casper - going to keep it simple to start, with the slide-along menu here http://tympanus.net/Development/SidebarTransitions/

Initially this can be triggered from the blog-logo.

@PaulAdamDavis mind doing the first pass on this? I can't think of any special considerations or additional notes at the moment.

PaulAdamDavis:
Closes #173

Navigation that pops out from the side.

nav-new

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -1,4 +1,4 @@
-/* ==========================================================================
+ ==========================================================================
    Table of Contents
    ========================================================================== */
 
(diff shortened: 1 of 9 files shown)


nuclearpengy:
Just thought I would comment on this and say how cool it works when you remove all navigation items and the menu just disappears from the front end. Zero effort at all. Freaking elegant. :heart: It!

One thing I did notice though(which is not critical), when the menu is removed/hidden, the "subscribe" button is no longer visible.

A cool upgrade might be to have something like: if no menu items show 'subscribe button' else load menu.

Thanks again. :)

PR #182, opened 2015-03-04
Prefixed second instance of tt with pre

mankittens:
It's not exactly clear, I could be mistaken, but it seems like the second instance of `tt` should be `pre tt`, otherwise the first set of `tt` rules are rendered mostly useless.


JohnONolan:
Nice catch

PR #189, opened 2015-03-23
Remove some !important declarations

PaulAdamDavis:
Relates to b071548

- Fixes an issue where the menu button would be the wrong colour

The navigation was using `!important` in a few places. I've changed the CSS a bit to no longer need these, but that means I've made some selectors slightly more specific.

The use of `!important` was to get around the hover state being persisted when the menu was opened but the mouse hadn't moved, which meant the menu button kept its white background when it should be black.
Making the hover state only apply when the nav is closed, gets around that.

I also added comments to a few other `!important`'s

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -689,7 +689,7 @@ body.nav-opened .nav {
     right: 30px;
     height: 38px;
     padding: 0 20px;
-    color: #111 !important;
+    color: #111 !important; /* Overides `.nav a:link, .nav a:visited` colour */
     text-align: center;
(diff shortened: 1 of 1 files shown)


PaulAdamDavis:
Updated to fix tag pages and so on.

ErisDS:
Does this fix the issue where the icon appears white by default?

PaulAdamDavis:
Yep :) I've updated the commit & PR message to clarify that.

screen shot 2015-03-31 at 10 27 58

PR #191, opened 2015-03-31
Display Subscribe button if Navigation Menu is not set

keiga-san:
Display 'Subscribe' button on every page if the 'Navigation Menu' is not set.

Class for standalone subscribe button is set to 'subscribe-button-standalone'. 
Button CSS style is taken from the previous version of theme.

subscribe_button

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -776,6 +776,32 @@ body.nav-opened .nav {
     width: auto;
 }
 
+.back-button {
+    -webkit-box-sizing: border-box;
+    -moz-box-sizing: border-box;
+    box-sizing: border-box;
+    display: inline-block;
+    float: left;
+    height: 38px;
+    padding: 0 15px 0 10px;
+    border: transparent 1px solid;
+    color: #9EABB3;
+    text-align: center;
+    font-size: 12px;
+    text-transform: uppercase;
+    line-height: 35px;
+    border-radius: 3px;
+    background: rgba(0,0,0,0.1);
+    transition: all ease 0.3s;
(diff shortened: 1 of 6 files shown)


keiga-san:
Add code for "Home" button. 

Now if Navigation Menu is not set, theme will look like it was in previous versions: with just "home" and "subscribe" buttons.

back_button

nuclearpengy:
my two cents worth: If a user is already on the "home page" then I think the home button should be removed. Would need to clear this with @JohnONolan first though.

Take a look at how @JohnONolan's site looks, I think this is a good natural behaviour.

home page:

screen shot 2015-04-01 at 23 40 03

Post page:

screen shot 2015-04-01 at 23 41 01

keiga-san:
@nuclearpengy Yes , you're right. I did so already. "Home" button appears only on: page, tag, author and post.

nuclearpengy:
oh legend. that's awesome. thank you. made my evening. I'm going to go drink coffee and celebrate ;)

NerosTie:
IMHO, the "HOME" button should be always here (except on the homepage), even if "menu" is set.
Just like before.

keiga-san:
@NerosTie yes, thats right. Now "Home" button appears on the left if there are no logo is set.
On the right you can see "Subscribe" if there are no Menu, or "Menu" button if it set.

NerosTie:
@keiga-san Cool! I hope it will be in the next update! :smile:

MichielDeMey:
:+1: Care to solve the merge conflicts? :smile:

Issue #190, opened 2015-03-31
Return the old one "Subscribe" button to default theme.


keiga-san:
Hello.

After update to new version with `{{navigation}}` there are no more cool hovered "Subscribe" button on the page. Now it's located in the navigation menu. So if i don't need a navigation menu there will be no subscribtion button.

Can you please return it to code if there are no navigation menu is set?

JohnONolan:
Sounds like a sensible suggestion - PR welcome

keiga-san:
Tried to make one.
#191

JohnONolan:
Will come back to this in future

PR #198, opened 2015-04-17
Subscribe and back buttons, part 2

JenniferMack:
re: https://github.com/TryGhost/Casper/issues/190

I updated the home/back button code against the latest commit on master, this is sort of a duplicate of https://github.com/TryGhost/Casper/pull/191, but it merges cleanly. I didn’t know how to fix it against the other PR, this seemed easier.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -781,6 +781,56 @@ body.nav-opened .nav {
     width: auto;
 }
 
+.back-button {
+    -webkit-box-sizing: border-box;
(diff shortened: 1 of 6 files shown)


PaulAdamDavis:
Hi @JenniferMack, I don't think this is supposed to be here? Closing just in case :)

PaulAdamDavis:
Ahhh, my mistake. Reopening.

MichielDeMey:
It seems both PRs have merge conflict now. ;)

PR #200, opened 2015-04-19
Social media icons with brand colors on hover

HLFH:
Hi,

According to http://brandcolors.net/,

IMHO, the "very dark grayish blue" color on hover of the social media icons is not fantastic and it could be a better secondary call-to-action by using the official brand colors on hover:
- #55acee for Twitter
- #3b5998 for Facebook
- #dd4b39 for Google+

You could see the difference:
social-media-icons


JohnONolan:
Good call :+1:

PR #202, opened 2015-04-23
Disable horizontal scrollbar in post page

minwe:

PR #213, opened 2015-06-22
Fixed horizontal scroll on the post page

lukaszklis:
I've spotted that on the post view there is a horizontal scroll present and I've started debugging what's the cause. The post has a previous one and a next one available, therefor you can see links to them in the bottom of the page. It turns out, the previous post link has an additional 1px border, which makes the element to have 50% + 1px  of width and that's why you can see a horizontal scroll (see screenshot).

Horizontal scroll caused by an additional border

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -1441,6 +1441,8 @@ body:not(.post-template) .post-title {
 /* if there are two posts without covers, put a border between them */
 .read-next-story.no-cover + .read-next-story.no-cover {
     border-left: rgba(0,0,100,0.04) 1px solid;
+    -webkit-box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


lukaszklis:
Fixes #202

JohnONolan:
Thanks for this - would you mind adding the appropriate prefixed versions, too?

lukaszklis:
Not at all, which browser do we target? As far as I can see, there might be a need for `-webkit-` prefix according to caniuse: http://caniuse.com/#search=box-sizing

JohnONolan:
-webkit- and -moz- please :)

lukaszklis:
@JohnONolan there you go. :)

PR #217, opened 2015-08-10
Added data attribute with post UUID to article element in post template

ryexley:
Having the Post UUID available in the template will make it available for use by code injection scripts that can make valuable use of it. Specifically, [I've created one](https://gist.github.com/ryexley/4e7ad06d9ce5f9014423) that injects Disqus comments on post pages that could really benefit from having the post UUID available for use in the `disqus_identifier` variable to allow Disqus to uniquely identify the post it associates comments with.

### DIFF
--- a/post.hbs
+++ b/post.hbs
@@ -16,7 +16,7 @@
 </header>
 
 <main class="content" role="main">
-    <article class="{{post_class}}">
+    <article class="{{post_class}}" data-post-id="{{uuid}}">
 
         <header class="post-header">
             <h1 class="post-title">{{title}}</h1>


JohnONolan:
This is a great implementation of UUID, with a valid use-case. Unfortunately it fails the most important test for a default theme which is designed to help people understand Ghost themes:

> Will it be useful to most people? Can this be added without causing confusion for most people?

In this instance it's a feature which is caters only to fairly advanced use-cases, and would be more than likely to confuse beginner developers who open the Casper theme for the first time and are already very overwhelmed.

We'll probably make use of this exact implementation in several future (more advanced) themes, but just not this one :)

PR #222, opened 2015-10-19
Move jQuery before `{{ghost_foot}}`

ErisDS:
- Including jQuery before `{{ghost_foot}}` means jQuery can be used inside ghost_foot code injection
- It should also improve backwards compat, as without this jQuery code included via ghost_foot may break

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -37,10 +37,11 @@
 
     </div>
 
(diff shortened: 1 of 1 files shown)


kevinansfield:
:+1: sounds like a good move to me

PR #226, opened 2016-01-06
Added mathjax and syntax highlighting integrations

shakedlokits:
As more and more people are using it, I find it's a good idea to add in the 
mathjax and syntax highlighting integrations into casper.
It would also solve the issues recurring with people who try to integrate these fixes while hosting
ghost on Heroku and other deployment PAAS.


JohnONolan:
You can add it here without needing to edit your theme, and have it still be there even if you change theme :) http://support.ghost.org/use-code-injection/

PR #227, opened 2016-01-07
Disable discretionary ligatures in headings

mecab:
Hi, I propose to disable `discretionary ligatures` in heading style because of following reasons: 

1. `Open Sans` is used for heading font face, but the font does not have discretionary ligatures AFAIK, though it has usual ligatures (`liga`).

2. For the letters other than English, it will be default to their `sans-serif` fonts, and the ligature specification might cause unexpected concatenations. Here's some funny Japanese examples:

 **Expected**: Word for "build"  
 Build

 **Failed**: Consolidated character for "building" (used in media with limited space e.g. newspapers) + sound "do"  
 Building + do

 **Expected**: Word for "france"  
 France

 **Failed**: The set for "franc" + sound "su"  
 Franc + su
 
 I am not sure if such a things happen in other languages, but discretionary ligatures is discretional, so I think it should not be specified in default except the font covers (almost) all language and we know how it combines the characters.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -206,9 +206,9 @@ body {
 
 h1, h2, h3,
 h4, h5, h6 {
-    -webkit-font-feature-settings: 'dlig' 1, 'liga' 1, 'lnum' 1, 'kern' 1;
-    -moz-font-feature-settings: 'dlig' 1, 'liga' 1, 'lnum' 1, 'kern' 1;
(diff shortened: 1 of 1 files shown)


JohnONolan:
Fork of the theme is the correct solution for now, in future we'd definitely like to have comprehensive support for i18n

mecab:
Thank you for your response, I'll fork the project for now.

PR #228, opened 2016-01-20
Disable ligatures in code blocks

jegtnes:
I recently created a new Ghost blog and a colleague pointed out that ligatures for code blocks looked a bit off due to the monospaced font. They weren't enabled in Firefox by default, but in Chrome (possibly different browser defaults?). This PR standardises it and removes any ligatures from `<code>` and `<tt>` blocks.

Before:
A block of code with ligatures enabled

After:
A block of code with ligatures disabled


JohnONolan:
Nice :+1:

PR #229, opened 2016-01-21
Update libraries to lastest version and minor fix

FongWan:
# screen.css
+ Update 'normalize.css' to v3.0.3
+ Lowercase all hex color to match the general style
+ Add indent to some properties with vendor prefix to the match general style
+ Fix all value of properties related to border to match formal syntax (`<br-width> || <br-style> || <color>`)
+ Remove redundant `top: 0` from `.nav-close:after`
+ Change `bold` from `font-weight` property to `700`
+ Reorder properties, grouped by type (position, display & box model, color, text and other)

# default.hbs
+ Change illegal value `|` to `%7c` at Google Fonts link to be represented in "Normalization Form C" specification
+ Remove all closing slash on void elements
+ Remove useless `type="text/css"` and `type="text/javascript"`
+ Use CloudFlare CDN to fetch JQuery and FitVids
+ Update JQuery to v1.12.0
+ Move `{{ghost_foot}}` after all javascript that are used by default.

# author.hbs, index.hbs, page.hbs, post.hbs and tag.hbs
+ Remove closing slash on void element
+ Remove `role` property on `<main>` element
+ Changed some `<section>` to `<div>` to avoid W3C validator warning when inside miss header.

# loop.hbs
+ Change `<section class="post-excerpt">` to `<div class="post-excerpt">` to avoid W3C validator warning when the content miss header.
+ Change author thumb image display by using `<div>` with background instead `` to avoid W3C validator warning with `nopin` property.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -22,82 +22,80 @@
 */
 
 /* ==========================================================================
-   0. Normalize.css v2.1.3 | MIT License | git.io/normalize | (minified)
+   0. normalize.css v3.0.3 | MIT License | git.io/normalize | (minified)
    ========================================================================== */
 
-article, aside, details,
-figcaption, figure,
-footer, header, hgroup,
-main, nav, section,
-summary { display: block; }
-audio, canvas, video { display: inline-block; }
-audio:not([controls]) { display: none; height: 0; }
-[hidden], template { display: none; }
-html {
-   font-family: sans-serif;
-   -ms-text-size-adjust: 100%;
-   -webkit-text-size-adjust: 100%;
+html { 
+    font-family: sans-serif;
+    -ms-text-size-adjust: 100%;
+    -webkit-text-size-adjust: 100%
 }
 body { margin: 0; }
-a { background: transparent; }
-a:focus { outline: thin dotted; }
-a:active, a:hover { outline: 0; }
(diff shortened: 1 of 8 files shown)


JohnONolan:
Hey there. Thanks for getting involved and helping to improve Casper! Lots of this is good, but there are some parts which aren't. Generally it's better to do smaller PR's tackling one thing at a time so that good things can be merged individually.

- We deliberately load local assets rather than remote by default - this is for privacy.
- Font-weight declaration is deliberate and intended
- CSS property re-order sounds correct from you description, but hard to tell because so much has changed. Generally wouldn't consider merging this unless it was generated by the same CSSComb rules used in Ghost core
- Self-closing tags are valid in HTML5, and used deliberately - as is out preference
- The position of ghost_foot is deliberate and intended
- W3C validator shows warnings, not errors, for sections without heads. This is not an issue / or something we want to change.
- role attributes are very much deliberate to aid with accessibility 
- nopin attribute is deliberate and intended

If you want to break the rest down into smaller PRs - with 1 descriptive commit per piece of work - I'd be happy to merge

FongWan:
Ok, sound good. I am new on this and did not know the rules and recomendations, just wanted to help. Let me do this again. :+1:

PR #234, opened 2016-01-23
Fix border syntax order on 'screen.css'

FongWan:
+ Fix all value of properties related to border to match formal syntax `<br-width> || <br-style> || <color>` (See: https://developer.mozilla.org/en/docs/Web/CSS/border)

When the syntax order go wrong, some CSS minifier process the file in unexpected way.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -297,7 +297,7 @@ hr {
     display: block;
     height: 1px;
     border: 0;
-    border-top: #EFEFEF 1px solid;
+    border-top: 1px solid #EFEFEF;
     margin: 3.2em 0;
     padding: 0;
 }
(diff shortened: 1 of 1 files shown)


JohnONolan:
PRs need to be rebased onto latest master before being submitted -- but border property order isn't something that we're super passionate about.

FongWan:
This PR is based onto latest master (that is why I pulled the master into my folk). When I tried some CSS minifier and all of them process the border property in unexpected way, like deleted border width, deleted border color, etc. The fix of the value order can fix all this thing, please reconsider.

JohnONolan:
You need to pull master before you make your changes, not after, so the history is clean. 

We're not going to change the order of border CSS properties. I would suggest reporting the behaviour you experienced as a bug to the CSS minifier.

PR #236, opened 2016-01-25
Fix minor difference between official normalize.css

hwdsl2:


hwdsl2:
There is a minor discrepancy in the normalize.css 3.0.3 patch by @FongWan (Ref: 591d1fb and 7f400e4) when compared to the official version at https://github.com/necolas/normalize.css/

This pull request fixes the discrepancy on the relevant line.

Also, please note that @FongWan 's patch also changed font weight "700" on some lines back to "bold" in the official normalize.css. I assue that is not an issue?

FongWan:
See: https://github.com/necolas/normalize.css/blob/master/normalize.css#L376

There are discrepancy between github.com version and github.io version. In the "official" version you linked there are not `box-sizing: content-box;` since [@patrickhlauke commit at Jan 6 (`fe56763`)](https://github.com/necolas/normalize.css/commit/fe567632afcf8afcf17925ab2a507ed87504cce8)

hwdsl2:
Ah I see - I was looking at the 3.0.3 release of normalize.css, which does not contain this change which was committed after the release.

Thanks @FongWan. Closing this pull request.

PR #240, opened 2016-02-20
Add fallbacks to OS-provided monospace fonts

mecab:
Hi, I found `Inconsolata` is specified to `code`, `tt` and `pre`, however it is not native font for major OSs (i.e, Windows, Mac). This PR provides fallbacks to OS native monospace fonts (`Consolas` for Windows and `Monaco` for Mac), which are similar to `Inconsolata`.

`Inconsolata` is also available in Google Fonts, so I think loading it as a web font is another option, while I did not do it in this PR.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -336,7 +336,7 @@ mark {
 
 code, tt {
     padding: 1px 3px;
-    font-family: Inconsolata, monospace, sans-serif;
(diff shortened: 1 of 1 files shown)


JohnONolan:
The intended native fallback (for now) is `monospace` which is usually the most appropriate for any given platform. Will probably circle back to this in future for another review when Casper next gets a design update though!

PR #243, opened 2016-02-28
Hide home page content before animating it in

stevenmirabito:
The home page title and description are not hidden when the page is loaded, so for a split second before the CSS animation kicks in you can see both elements on the page. This makes the animation look choppy and late:

This small tweak sets the opacity of those elements to zero by default, then adds the the `forwards` fill mode to the animation so the last keyframe will stay on the page once the animation completes. This is the result:

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -901,16 +901,22 @@ body.nav-opened .nav {
 
 /* Add subtle load-in animation for content on the home page */
 .home-template .page-title {
+    opacity: 0;
     -webkit-animation: fade-in-down 0.6s;
     animation: fade-in-down 0.6s;
     -webkit-animation-delay: 0.2s;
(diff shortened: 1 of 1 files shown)


sawilde:
well I liked it and added it to my fork :)

eibt:
By setting animation-fill-mode to "both" instead of "forwards" you don't need to set the opacity to 0 so you save 2 lines of code.

Even better, you can define the fill mode in the animation property so you save another 4 lines:

`-webkit-animation: fade-in 0.6s both;`
   ` animation: fade-in 0.6s both;`

As suggested by @matthias-vogt here https://github.com/TryGhost/Casper/pull/268

JohnONolan:
Valid bug, but more succinct fix merged from https://github.com/TryGhost/Casper/pull/268

PR #246, opened 2016-03-20
Fixed issue with aspect ratio of author thumbnail in posts loop

### LINKED ISSUE #237: .author-thumb lacks object-fit or similar
The author thumbs which display on the index page gets squashed or stretched if the author profile image isn't perfectly square. Fixed it for myself by setting object-fit: cover to the css doc, on line 974.

allistercsmith:
closes #237
- changed author image in loop partial from 'img' tag to 'a' tag, with a background-image style
- changed the author-thumb CSS class to center background and set it to cover

### DIFF
--- a/partials/loop.hbs
+++ b/partials/loop.hbs
@@ -13,7 +13,7 @@
         <p>{{excerpt words="26"}} <a class="read-more" href="{{url}}">&raquo;</a></p>
     </section>
     <footer class="post-meta">
-        {{#if author.image}}<img class="author-thumb" src="{{author.image}}" alt="{{author.name}}" nopin="nopin" />{{/if}}
+        {{#if author.image}}<a class="author-thumb" style="background-image: url({{author.image}})" nopin="nopin" />{{/if}}
         {{author}}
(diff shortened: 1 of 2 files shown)


partials/loop.hbs
```diff
@@ -13,7 +13,7 @@
         <p>{{excerpt words="26"}} <a class="read-more" href="{{url}}">&raquo;</a></p>
     </section>
     <footer class="post-meta">
-        {{#if author.image}}<img class="author-thumb" src="{{author.image}}" alt="{{author.name}}" nopin="nopin" />{{/if}}
+        {{#if author.image}}<a class="author-thumb" style="background-image: url({{author.image}})" nopin="nopin" />{{/if}}
```
kevinansfield: This is invalid HTML because `<a>` elements aren't void-elements and so can't be self-closing.

@JohnONolan you mentioned in https://github.com/TryGhost/Casper/issues/237#issuecomment-179026938 that it needs to be an `a` tag, could you elaborate on the reasoning?

PR #251, opened 2016-05-22
Fixing Table of Contents in CSS

max-nova:
Noticed that the CSS table of contents got out of sync with the code.  Fixed!


acburdine:
Thanks!

PR #254, opened 2016-06-26
Fix invalid css property declaration

sergioaugrod:
Fix invalid CSS property declaration.


acburdine:
@sergioaugrod so I know those are invalid comment strings, but imo it would be better to rephrase the comment (and use the correct comment delimiters) so that the fact that it is a hack isn't lost. :smile:

sergioaugrod:
@acburdine You're right. I changed the message and put the correct comment delimiters.

acburdine:
@sergioaugrod thanks!

PR #258, opened 2016-07-15
update from upstream

choicky:


acburdine:
I don't think this belongs here?

@choicky if you can provide a reason for these changes we can reopen this. Otherwise, I think this might have been an accidental PR upstream?

choicky:
Sorry.

I intended to update my repo according the base...

PR #262, opened 2016-07-28
No margin-bottom for last p inside li tags

### LINKED ISSUE #241: Nested list showing sublist too close to next main list item
Consider this markup simulating a definition list:

``` MARKDOWN
Here we experiment with Ghost.

- **One**  
  The first list item, with sublist.

  - Sub list item A;
  - Sub list item B;

- **Two**  
  Third main item.

- **Three**  
  The final item on this list.
```

Note that (a) there are *two spaces* after each primary list item's first line, and (b) that the behavior doesn't change if I remove the whiteline between the sublist and the text above that.

This is rendered by Casper 1.2.8 like this (red pen by me):

----

casper

----

The same holds for the preview when editing.

To my taste there should be different margins, making the sub-list closer to its parent instead of close to its parent's next sibling.

jeroenheijmans:
Fixes #241.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -295,6 +295,10 @@ li li {
     margin: 0;
 }
 
+li > p:last-of-type {
+    margin-bottom: 0;
+}
+
 hr {
     display: block;
     height: 1px;


JohnONolan:
Nice fix - sorry about the delay

PR #267, opened 2016-08-28
Changes for reddit share

wdherndon:
Added "reddit share". The most difficult part was adding the reddit icon to the font, so I added a readme.html in the fonts folder about how I made the font changes. If you would like the documentation in a different location/form, I'd be happy to accommodate.


wdherndon:
Added LinkedIn, StumbleUpon and Email shares.

PR #272, opened 2016-10-07
Remove old body classes

AileenCGN:
refs TryGhost/Ghost#2597

Make Casper work without old body classes `.archive-template`, `.page` and use `.page-template` *only* for pages, as `.post-template` will not be available for pages anymore.


ErisDS:
LGTM 👍

PR #274, opened 2016-10-13
Use protocol-relative URL for jQuery script

eibt:
So after a while of not using Ghost I upgraded from 0.9.0 to 0.11.2, then I noticed the menu didn't work, because for some reason "https://code.jquery.com/jquery-1.12.0.min.js" was not resolving:

jquery-https

So I tried "http://code.jquery.com/jquery-1.12.0.min.js" and it was working perfectly, same with "//code.jquery.com/jquery-1.12.0.min.js" since my blog doesn't use SSL .

I see that the link to fonts.googleapis.com is a protocol-relative URL, so it would be good to do the same with code.jquery.com?

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -42,7 +42,7 @@
     </div>
 
     {{!-- jQuery needs to come before `{{ghost_foot}}` so that jQuery can be used in code injection --}}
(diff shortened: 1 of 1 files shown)


ErisDS:
Seems like a sensible & consistent change to make, thanks 👍

PR #275, opened 2016-10-14
Change wording

jwasham:
I changed inappropriate wording that bugs me every time I make additions to the stylesheet. 

Thank you for your hard work. I love this project!

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -519,7 +519,7 @@ button {
    3. Utilities - These things get used a lot
    ========================================================================== */
 
-/* Clears shit */
+/* Clears elements */
 .clearfix:before,
 .clearfix:after {
     content: " ";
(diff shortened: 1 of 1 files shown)


ErisDS:
You're not [the first](https://github.com/TryGhost/Casper/pull/146), I doubt you'll be the last. As a blogging platform, we're all about freedom of speech. This kind of censorship / tone-policing flies in the face of our beliefs. Here's some [further reading](https://medium.com/@girlziplocked/skullfuck-you-very-much-cultural-authority-and-the-dispossessed-1770fdb01dc4#.p19hew26f) that explains why PRs like this do not get a warm reception.

jwasham:
I'm disappointed, but I understand your protection of free speech. 

Ghost is not a pet project. It's a widely used platform. To me, it's just about being respectful of the end user and being professional in your code. If you wouldn't allow that wording on your homepage, why would you allow it in your code?

I'll drop the issue and just deal with it. Thanks for hearing me out.

PR #279, opened 2016-11-30
move font above css include

Humeira:
I noticed fonts are included just after the css, I think it's better to include it before to avoid FOUC.


JohnONolan:
Excellent point, sorry for missing this PR!

PR #295, opened 2017-03-17
White space mobile

sarandford:
This is my first time committing, not sure why the default.hbs file is in there and how that happened but it should still work. Let me know what I can do to improve it!

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -2082,7 +2082,7 @@ body.page-template .post-title {
     }
 
     .post-content img {
-        padding: 0;
+        padding: .06em;
         width: calc(100% + 32px); /* expand with to image + margins */
         min-width: 0;
         max-width: 112%; /* fallback when calc doesn't work */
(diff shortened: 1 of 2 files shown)


ErisDS:
Hi @sarandford congrats on your first commit! This is not a comment on whether or not the CSS change would be approved - 'fraid that's not my area, but I can provide some feedback on the PR itself.

It seems that when you created your `white-space-mobile` branch, you created it from our `ghost-1.0` branch, which is a branch we're using for updating Casper with backwards incompatible changes ready for when Ghost 1.0 is shipped.

However, when you made the PR, you made it between your branch, and Casper's `master` branch, this can be seen at the top of the PR:



Because the branch you started from had extra commits that our master branch doesn't have, your PR has scooped those up and included them. We cannot merge this, because it would add the breaking changes onto to the master branch and be included in Ghost 0.11.8. 

As a general rule you need to make your PR *to* the branch you started *from* - so if you hit the "Edit" button that appears in the screenshot above, you can change the base branch to ghost-1.0 instead of master. This is the easiest solution.

We can merge the PR for the Ghost 1.0 branch, and also copy it to the master branch.

sarandford:
Thanks so much @ErisDS, closing this and submitting a different one

PR #298, opened 2017-03-23
radio buttons and checkboxes appear regularly on chrome and safari

margemcm:
because  `-webkit-appearance: none;` messes with the display of radio buttons and checkboxes in chrome and safari, this styling now applies to all other input types, but not radio buttons or checkboxes.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -446,7 +446,11 @@ margin on the iframe, cause it breaks stuff. */
     margin: 0;
 }
 
-textarea, select, input {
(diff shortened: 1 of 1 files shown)


spaceninja:
A suggestion: Rather than target every text input, try using the :not() selector to make it more clear that what you're doing is excluding radios and checkboxes from your selection.

```css
input:not([type="radio"]):not([type="checkbox"])
```

margemcm:
thanks!  I'll fix it up and resubmit.

PR #310, opened 2017-06-03
Allow line breaks from `Bio`

PhoenixPeca:
I think we should give the authors freedom to have line-breaks in their bio.
_________________________

Bio with line-break in the admin panel:

Bio with line-break, without `white-space: pre;` CSS directive:

Bio with line-break, with `white-space: pre;` CSS directive:

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -1392,6 +1392,7 @@ body.page-template .post-title {
     color: #50585D;
     letter-spacing: 0;
     text-indent: 0;
+    white-space: pre;
 }
 
 .author-meta {


JohnONolan:
Nice 👍

PhoenixPeca:
Hi @JohnONolan,

Just an update, we can have this instead:
```
 white-space: pre-wrap;       /* css-3 */
 white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
 white-space: -pre-wrap;      /* Opera 4-6 */
 white-space: -o-pre-wrap;    /* Opera 7 */
 word-wrap: break-word;       /* Internet Explorer 5.5+ */
```
This will  allow users to have line breaks as their will and wraps text if it gets too long. 😉 

source: https://css-tricks.com/snippets/css/make-pre-text-wrap/

PR #313, opened 2017-06-13
[WIP] Casper 2.0

JohnONolan:
Almost done, just opening this PR so anyone who wants to play with it early - can.

image


jloh:
This looks awesome! Great work. The menu styling doesn't seem to work in this branch yet, but I understand its a WIP!

vikaspotluri123:
+1 @jloh!
To expand what he is saying, when running casper2 in chrome (and most likely other browsers), the menu is rendered as a list, with bullets and line breaks, instead of a horizontal "menu", like the design would suggest.

default.hbs
```diff
@@ -1,55 +1,57 @@
...
     <meta name="HandheldFriendly" content="True" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
-    {{!-- Brand icon --}}
-    <link rel="shortcut icon" href="{{asset "favicon.ico"}}">
-
     {{!-- Styles'n'Scripts --}}
-    <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Merriweather:300,700,700italic,300italic|Open+Sans:700,400" />
-    <link rel="stylesheet" type="text/css" href="{{asset "css/screen.css"}}" />
+    <link rel="stylesheet" type="text/css" href="{{asset "assets/public/screen.css"}}" />
```
jloh: Does the asset helper change in 1.0 or does this have an extra `assets` in it?

kevinansfield: The leading `assets/` is optional - if it's omitted then the `{{asset}}` helper will add it for you except for two special cases `favicon.ico/png` and `public/*`

jloh: Ah right, cool! Thanks for replying.

JohnONolan: @kevinansfield FWIW that is not the case at the moment, at least on Ghost 1.0. When leading `assets/` is omitted, it just breaks - asset helper doesn't fix it.

kevinansfield: In this case it wouldn't work because `public/` is also a special case or am I misunderstanding what you mean by breaks?

JohnONolan: ah no - I get it now - fixed

PR #316, opened 2017-06-21
💄  Infinite scroll

### LINKED ISSUE #314: Casper 2.0 infinite scroll
I almost forgot - the one thing missing from Casper 2.0 is that it could really do with some infinite scrolling. Right now I've just set `posts_per_page` to 25 so that there's quite a lot content on screen at a time, but really we want this to be autoloading new posts as you scroll down, 25 at a time is probably about right.

AileenCGN:
closes #314

Adds infinite scroll logic to Casper 2.0

This solution needs `$.get` which is not included in `jquery-3.2.1.slim.min.js`. Maybe there's another way to do this?

Here's a first glimpse (`config.posts_per_page` set to 6 for testing):

infinite_scroll

### DIFF
--- a/assets/js/infinitescroll.js
+++ b/assets/js/infinitescroll.js
@@ -1,0 +1,18 @@
+// Code snippet inspired by https://github.com/douglasrodrigues5/ghost-blog-infinite-scroll
+$().ready(function () {
+    var page = 2,
+        blogUrl = window.location,
+        $result = $('.post-feed');
+
+    $(window).scroll(function () {
+        if ($(window).scrollTop() + $(window).height() == $(document).height()) {
+            if (page <= maxPages) {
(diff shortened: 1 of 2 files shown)


AileenCGN:
Two things here, that I'm not sure about and maybe @JohnONolan and/or @ErisDS can help me here:

1. Can you think of a way to do the request via the Ghost API here? I tried it, but as it's returning a `json` object and I wasn't able to find a way to render it properly.
2. I couldn't find a way to just load the `$.ajax` / `$.get` functionality from jquery, because it's not included in the slim version. So I needed to load the normal minified version of it. Can someone think of a better solution?

vikaspotluri123:
@AileenCGN What about the [alternative](http://youmightnotneedjquery.com/#request)? Is there anything preventing the use of it?

PR #319, opened 2017-06-25
Remove infinitescroll in single page

zce:
Remove infinitescroll in single page


zce:
if url has params or hash, it does not work

https://github.com/TryGhost/Casper/blob/master/assets/js/infinitescroll.js#L4

AileenCGN:
Thanks for the improvement, @zce !!! 👍

PR #334, opened 2017-07-29
Prevent floating navigation from blocking the text header in fragmented urls

PhoenixPeca:
`partials/floating-header.hbs` blocks the H1-H6 header texts in fragmented urls. So, as a workaround, we will need to delegate a few space above the header.

**Without** this PR, the page would look something like this:
image

---

**With** this PR, the page would look something like this:
image

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -667,6 +667,11 @@ The first (most recent) post in the list is styled to be bigger than the others 
     min-width: 100%;
 }
 
+h1, h2, h3, h4, h5, h6 {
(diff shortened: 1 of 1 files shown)


JohnONolan:
Nice fix 👍  - if you can make this specific to `.post-full-content` rather than global, just below line 668 then would be happy to merge

PhoenixPeca:
@JohnONolan done! 😉

JohnONolan:
haha, and you found a more sane place for it, too! Thanks! 🥇

PR #337, opened 2017-07-31
✨ Update to use Primary Tag syntax

ErisDS:
We're going to leave a week before merging this and shipping a minor bump of Casper.

Note: this also requires a documentation update

refs TryGhost/Ghost#8668

- Ghost 1.2.0 we introduce primary_tag as a calculated property
- Casper can be updated to use this for clarity
- We want to leave a short period before merging this, to minimise the risk of compatibility issues

### DIFF
--- a/partials/post-card.hbs
+++ b/partials/post-card.hbs
@@ -7,8 +7,8 @@
     <div class="post-card-content">
         <a class="post-card-content-link" href="{{url}}">
             <header class="post-card-header">
-                {{#if tags}}
-                    <span class="post-card-tags">{{tags.[0].name}}</span>
+                {{#if primary_tag}}
+                    <span class="post-card-tags">{{primary_tag.name}}</span>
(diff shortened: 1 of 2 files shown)


ErisDS:
This needs a second pass cos the syntax can be better 😁

ErisDS:
I have updated this PR, it is ready for review. Might need to think about versioning around this change 🤔 

Note: there is also no documentation in themes.ghost.org/docs/primary_tag yet.

partials/post-card.hbs
```diff
@@ -8,7 +8,7 @@
         <a class="post-card-content-link" href="{{url}}">
             <header class="post-card-header">
                 {{#if tags}}
```
kirrg001: `{{#if primary_tag}}`

ErisDS: good catch fixed

Issue #349, opened 2017-08-10
Card issue after upgrading to Casper 2.0.3


JDevlieghere:
After upgrading to Ghost 1.5.2 and Casper 2.0.3 I noticed that the cards on my blog looked a little off. It looks like there is some kind of overlay, which makes the blog title greyed-out and  also prevents me from clicking on the post title. I tried to figure out what element was causing this with the Chrome's DevTools but without success. 

Returning to the previous version with `ghost update --rollback` resolved the issue. I am not injecting any custom CSS other than the snippet below. 

```
<style type="text/css">
	.post-full-content pre {
		background: #002b36;
		border: 1px solid #002b36;
	}
</style>
```

**After the upgrade:**
![after](http://i.imgur.com/qkPpkG7.png)

**Before (after the rollback):**
![before](http://i.imgur.com/NSorj8p.png)

Please let me know if there is any more info I can provide!

ifvictr:
Can confirm. This is my Ghost running with Casper 2.0.3: ![2017-08-10_15-41-50](https://user-images.githubusercontent.com/10110245/29195282-8285a89c-7de2-11e7-8420-78c09bcb4b55.png)

pascalandy:
Same here.

corazza:
Same!

PR #350, opened 2017-08-11
fixed z-index problem for cards on index page

nova0052:
z-index was causing cover image to overlap with first post card on index page. Changed z-index to 0 to fix. #349


ndaidong:
This may cause problem with the header on the post pages (not homepage). 
Safer fix may be:

```
.home-template .site-header:after,
.home-template .site-header:before {
  z-index: 0;
}
```
It just effects the homepage.

PR #360, opened 2017-08-15
Add progress bar styling for Firefox

gilrain:
FYI, this is my first open source contribution and my first pull request!

I alter the progress bar color to red for my own site, so it stood out much more easily than the Casper blue not matching the Firefox default. Autoprefixer isn't handling this one. Also, I tried adding this rule, comma separated, to the immediately above webkit rule... but it doesn't actually work when I do that, for some reason. Sorry for the redundancy!

https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-progress-bar


JohnONolan:
@gilrain Welcome! And thanks so much for honour of your first PR :) looks great :v:

PR #374, opened 2017-08-23
Add a 'build-dist' Gulp task to create the zip

YoranBrondsema:
Adds a Gulp task so that

````sh
gulp build-dist
````

builds the CSS and creates a `dist/<theme-name>.zip` archive. This makes uploading a custom theme based on Casper a little bit easier!

### CODE REVIEW

**Comment by JohnONolan** [comment]
Wow that big red X is a bit aggressive, I just meant to add preliminary approval & comment w/ change suggestion

File: gulpfile.js [hunk 1]
```diff
@@ -49,6 +50,22 @@ gulp.task('watch', function () {
     gulp.watch('assets/css/**', ['css']);
 });
 
+gulp.task('build-dist', ['css'], function() {
+    var rimraf = require('rimraf');
+    var fs = require('fs');
+
+    var targetDir = 'dist/';
+    var themeName = JSON.parse(fs.readFileSync('./package.json')).name;
```
ErisDS: This can be simplified to `themeName = require('./package.json').name;`

gulpfile.js
```diff
@@ -49,6 +50,22 @@ gulp.task('watch', function () {
...
+    var filename = themeName + '.zip';
+
+    // remove the old dist/ directory
+    rimraf.sync(targetDir);
```
ErisDS: This isn't needed - it will automatically overwrite 🙂

gulpfile.js
YoranBrondsema: Indeed, thanks!

ErisDS:
Late to the party here - I have drawn some inspiration from this PR whilst working on a similar tool to make it easier to deploy themes and wanted to feedback on a couple of small things I found. I've added 2 small comments, but I also believe there may need to be extra tweaks to ensure the css task finishes first - this may be the reason for the addition of rimraf, not sure.

I'm pretty new to gulp as well, and I read these two things, the first talks about using return: http://schickling.me/synchronous-tasks-gulp/ and the second `gulp.start`: https://markgoodyear.com/2014/01/getting-started-with-gulp/

gulpfile.js
YoranBrondsema: You're right, but I left it in case someone decides to change the name of the theme in between builds. This shouldn't occur too often but I thought it wouldn't hurt to start from a "clean slate" and cover this case as well. But if you prefer to have it removed, that's fine for me too!

ErisDS: My preference would be in favour of one-less-dependency. In this case a user can manually delete the file :)

YoranBrondsema: That's a good reason!

YoranBrondsema:
@JohnONolan Thanks for your comment. I renamed the task to `zip` and also added a little blurb to the README.

@ErisDS Thanks for your feedback! You had a good point about the synchronicity of the tasks. I wasn't aware of that either in Gulp. It seems that by adding `return` statements to the task functions, it properly serializes the task executions. It's also documented in the doc: https://github.com/gulpjs/gulp/blob/master/docs/API.md#return-a-stream.

ErisDS:
Awesome & fast work!  LGTM 👍

JohnONolan:
Sorry for being super late!

PR #375, opened 2017-08-25
Remove `border-radius` in `.post-card-image-link`

arniu:
Setting in `.post-card-image-link` will lead to the following result:

ghost


JohnONolan:
Thanks for reporting! This is a valid bug but not the right fix :) resolved in https://github.com/TryGhost/Casper/commit/269d2b57876b29852fd89a71b504c0cb0b0e8d2c

PR #381, opened 2017-09-04
[HOLD] 🔥  Remove `@labs.subscribers`

kirrg001:
no issue

Will be compatible with Ghost 1.9.

In Ghost 1.9 we enabled subscribers by default.
NOTE: `@labs.subscribers` is still available for themes to avoid breaking them.

Bump Ghost engine, because if you are using this version of casper < Ghost 1.9, you would not be able to disable the subscribers feature.

**I would wait with merging, because 1.9 will be released on Thursday and if there are any patch changes, we would have to revert this PR! That's why the PR is HOLD!**

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -41,7 +41,6 @@
     </div>
 
     {{!-- The big email subscribe modal content --}}
-    {{#if @labs.subscribers}}
     <div id="subscribe" class="subscribe-overlay">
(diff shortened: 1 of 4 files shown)


kirrg001:
This is super old and incorrect, because we haven't merged https://github.com/TryGhost/Ghost/pull/8954.

PR #395, opened 2017-10-01
chore: Add code hightlight

BoizZ:
Add hightlight in post


ErisDS:
Hi @BoizZ, I think this belongs in a fork? Syntax highlighting is something not every one needs, and therefore wouldn't be added to default Casper.

Meanwhile, it is possible to add it via code injection, without editing the theme.

PR #400, opened 2017-10-06
Removed double semicolons in CSS.

Scionar:
No issue. Just noticed two typos.


AileenCGN:
Thanks for the PR @Scionar!

Just one thing for future contributions: Could you run `gulp` before pushing the changes, so the `built` files are updated as well? 😜

PR #403, opened 2017-10-21
Optimize gulp zip file generation.

matitalatina:
I realized that `gulp zip` task contains two useless folders for production mode: `dist` and `css` folders.

- `dist` is not required because it contains the zip file itself.
- `css` is not required because template file uses `built` folder to get styles.

So I think it's better to strip out those folders to reduce zip size.


AileenCGN:
Thanks for the PR! LGTM 👍

PR #411, opened 2017-11-22
+noframe, -fitVids

### LINKED ISSUE #409: Replace Fitvids/jQuery with Reframe/Noframe
Anyone welcome to take this, if not I'll look at it later. 

https://github.com/dollarshaveclub/reframe.js

pankajpatel:
closes #409


kevinansfield:
Hey @pankajpatel 👋 Thanks for taking a look at this and sorry for no response for so long, there hasn't been a chance to review this in amongst everything else leading up the the christmas break.

Whilst waiting for a full review/test would you mind adding the [missing license](https://github.com/dollarshaveclub/reframe.js/blob/master/dist/noframe.min.js) to the `noframe.js` file?

kevinansfield:
The minified version of `noframe.js` should probably also be used as there is no other JS processing done in Casper to minify the file for distribution.

pankajpatel:
Alright!

pankajpatel:
@kevinansfield no worries about the response. PR updated!

AileenCGN:
Hey @pankajpatel 

Again, thanks for the PR!

I compared the behaviour of `noframe.js` vs. `fitvids.js` with some iFrames and we still need some style adjustments here.

See here with FitVids:
fitvids

Compared to noframe:
noframe

I didn't look any further into this, maybe you have an idea already?

pankajpatel:
@AileenCGN Yes this is caused due to the `noframe`'s styles addition which is `max-width: 100vw`

It will not work as there is already a padding on the container and more context aware setup is needed.

Initially I went ahead with `noframe` because it is only css based solution and once the iframe is touched, it will not be touched again by the `noframe` script.

But seems like `reframe.min.js` provides stable solution. Will commit the new solution with `reframe.js`

pankajpatel:
I will also squash the commits to one, don't wanna pollute the history with unnecessary commits.

AileenCGN:
Thanks for the update! Unfortunately this doesn't seem to work in Firefox now 😔

pankajpatel:
The only reason I have figured out till now is the `display: flex` on `.kg-card-markdown` which causes the child `.js-reframe` to not capture the needed height for itself on firefox.

I am not sure about why `display: flex` has been used on `.kg-card-markdown` as it is not present in any of the template. And the immediate parent of `.kg-card-markdown`, i.e. `.post-full-content` is `display: block`

Any suggestions? @AileenCGN @kevinansfield

pankajpatel:
Shall I delete this PR? As to make it work, other critical changes will be needed.

AileenCGN:
@pankajpatel I wouldn't delete it. We can leave it open but put a "[HOLD]" tag in the title. Maybe someone wants to pick it up from here...

pankajpatel:
@AileenCGN sounds good. Though if I can get few answers to the above question regarding classes and styles, maybe I can finish it.

kevinansfield:
`.kg-card-markdown` is part of the editor output so you won't see it in the theme templates. I'll put this PR on my list to pull down and evaluate, I'm sure there will be a fix available for FF

pankajpatel:
@thomasvaeth this PR has solution related to that; `reframe.js` does so with JS & CSS and `noframe.js` does so with CSS

`noframe.js` was not working properly on Firefox, whereas `reframe.js` works fine

PR #417, opened 2018-01-03
Remove gulpfile & css in the dist file

zce:
These files should not appear in the dist, because they can be accessed by the user.

> e.g.
> https://demo.ghost.io/gulpfile.js
> https://demo.ghost.io/assets/css/screen.css

But it's not necessary.

https://github.com/TryGhost/Ghost/blob/master/core/server/web/middleware/static-theme.js#L9

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -58,6 +58,8 @@ gulp.task('zip', ['css'], function() {
 
     return gulp.src([
         '**',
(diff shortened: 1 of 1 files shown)


kevinansfield:
Hey @zce 👋 Both of those assets are needed in the dist so that theme downloads from within the admin result in a usable/buildable theme.

PR #418, opened 2018-01-04
Update Feedly URL and to HTTPS protocol

jochenkirstaetter:
Hi,

Given that the [W3C Link Checker](https://validator.w3.org/checklink) marks the current URL of using Feedly as a permanent redirect (HTTP 301) it should be relevant to update the anchor sources accordingly in the default template, too.

Line: 120 http://cloud.feedly.com/ redirected to https://cloud.feedly.com/
Status: 301 -> 200 OK
This is a permanent redirect. The link should be updated to point to the more recent URI.

Broken fragments:
http://cloud.feedly.com/#subscription/feed/https://jochen.kirstaetter.name/rss/ (line 120)

Thanks for consideration.

Regards, Jochen

### DIFF
--- a/author.hbs
+++ b/author.hbs
@@ -30,7 +30,7 @@
                 {{#if facebook}}
                     <a class="social-link social-link-fb" href="{{facebook_url}}" target="_blank" rel="noopener">{{> "icons/facebook"}}</a>
                 {{/if}}
-                <a class="social-link social-link-rss" href="http://cloud.feedly.com/#subscription/feed/{{url absolute="true"}}/rss/" target="_blank" rel="noopener">{{> "icons/rss"}}</a>
+                <a class="social-link social-link-rss" href="https://cloud.feedly.com/#subscription/feed/{{url absolute="true"}}/rss/" target="_blank" rel="noopener">{{> "icons/rss"}}</a>
             </div>
         </div>
     </div>
--- a/author.hbs
+++ b/author.hbs
@@ -30,7 +30,7 @@
                 {{#if facebook}}
                     <a class="social-link social-link-fb" href="{{facebook_url}}" target="_blank" rel="noopener">{{> "icons/facebook"}}</a>
                 {{/if}}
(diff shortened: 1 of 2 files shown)


AileenCGN:
Hey @jochenkirstaetter 

Thanks for the report.

I tested the W3C link checker that you referred to with the old and the new URL from your blog and I can't see any difference.

See the old result:
screencapture-validator-w3-org-checklink-1515401322030

Compared to the new result:
screencapture-validator-w3-org-checklink-1515401163158

Can you provide some more information, why it is relevant to change the URL?

Thanks!

jochenkirstaetter:
Yes, because I fixed my local installation already according to the output the validator.

AileenCGN:
I tested it with the changed URL as well as the URL it was before (also with other blogs) and couldn't see a difference to the URL you submitted in this PR.

jochenkirstaetter:
Hi @AileenCGN 

Okay, I think it's still not clear what I'm trying to say. Please use the following [W3C link checker result](https://validator.w3.org/checklink?uri=https%3A%2F%2Fwww.mscc.mu&hide_type=all&recursive=on&depth=3&check=Check).
This checks the URLs in a Ghost blog instead of the Feedly URL that you checked.

The Ghost template used on that domain is derived from Casper, and produces the following link for the RSS button: href="http://cloud.feedly.com/#subscription/feed/https://www.mscc.mu/rss/" - that generated href-value is outdated.
The W3C link checker result indicates the change from HTTP to HTTPS, too.

image

When you actually following the link, Feedly redirects you once more to the final URL using https://feedly.com/ instead of http://cloud.feedly.com/ 

The generated URL in the RSS button by Casper should read like so: href="https://feedly.com/i/subscription/feed/https://www.mscc.mu/rss/" instead. 
My PR simply updates the protocol and domain part of the RSS button code in the Casper template, as seen [here](https://github.com/TryGhost/Casper/pull/418/files).

Hope this helps.

Best regards, Jochen

HLFH:
@AileenCGN Hi. Can you please close #424 and merge #418 ? Thanks

AileenCGN:
@jochenkirstaetter Gotcha! Thanks for the further explanation!

jochenkirstaetter:
Hi @AileenCGN 
Awesome, thanks for accepting the PR.

PR #419, opened 2018-01-12
Only show Twitter/FB if they are populated

conorgil:
The floating header was showing the Facebook and Twitter social sharing icons even if the blog did not populate those fields in settings.

This PR adds the necessary conditionals to only show the sharing options which exist.

### DIFF
--- a/partials/floating-header.hbs
+++ b/partials/floating-header.hbs
@@ -11,14 +11,18 @@
     <div class="floating-header-title">{{title}}</div>
     <div class="floating-header-share">
         <div class="floating-header-share-label">Share this {{> "icons/point"}}</div>
+        {{#if @blog.twitter}}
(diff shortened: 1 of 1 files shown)


AileenCGN:
Hey @conorgil 👋 

Thanks for the PR! 

The floating social links in the floating header are not dependent on Twitter or Facebook being set up for the blog, as they are simply share buttons and only use the URL of the current post that shall be shared. The actual `@blog.twitter` and `@blog.facebook` properties are not used here, therefore I don't see the usage in making the share functionality dependent on them.

If you don't want to use those social media share buttons, then I suggest that you remove them in your theme, as I believe these social media share buttons are pretty useful for users in general 😉

conorgil:
AH! Well, that does make more sense. Thanks for being kind to a Ghost newbie :)

PR #425, opened 2018-01-27
Fix blog title margin

Angristan:
Before this PR:
screenshot_27-01-2018_12-18-55
As you can see the last character is a bit cropped.

After this PR:
screenshot_27-01-2018_12-18-47


JohnONolan:
I can see the 1px difference but adding a 1px margin is addressing the symptom rather than the cause. The correct fix for this would be to address why the character is being cut off (I don't know) and adjust the base code so that it no longer happens. 

No expectation for you to do another PR if you don't want to - just letting you know why I'm closing this one for now!

Angristan:
You're right indeed, it isn't a good way to solve it. :)

After a bit more research, I found the cause: https://github.com/TryGhost/Casper/blob/99fad4cbc430427b01510f0f5e06c9671bc64265/assets/css/screen.css#L1398

However I don't really know what to do from there, as modifying this property will change the entire title.

PR #438, opened 2018-02-28
Ditched jQuery.

thomasvaeth:
– Rewrote the infinite scroll script in vanilla JS
– Ditched FitVids for Reframe.js (not Noframe.js), which is issue #409


kevinansfield:
@thomasvaeth thanks for looking into this! Unfortunately dropping jQuery completely will be a major breaking change for a lot of sites so it will likely have to wait for a major version bump of Ghost before it can go in.

If you keep the jQuery include then the non-jquery infinite scroll and reframe.js can parts can be reviewed. Did you see the existing [reframe.js PR + discussion](https://github.com/TryGhost/Casper/pull/411)? There are issues with the styling when using reframe, does your PR address those?

thomasvaeth:
@kevinansfield Yeah. So that PR is using Noframe.js (which is in the same repo as Reframe.js, so confusing a little). Noframe doesn't wrap elements like Reframe and FitVids, but uses CSS calc to scale the elements. I can try to recreate the differences between Reframe and Noframe this evening, but I didn't see any differences between Reframe and FitVids.

kevinansfield:
@thomasvaeth that PR is using `reframe.js` - it started out using `noframe.js` but was switched to `reframe.js` and still has problems in Firefox due to Casper's use of flexbox.

There are two things that are required:
- [ ] a fix for reframe.js in FF
- [ ] adding the jQuery include back in this PR - it's a major breaking change so would need to be handled separately

thomasvaeth:
@kevinansfield I just added jQuery and fitVids back into my PR. I will jump on the other PR to figure out the Reframe/Noframe thing for FF.

kevinansfield:
@thomasvaeth it may be best to split this into two PRs, one for removing jQuery from the infinite scroll and one for changing the fitVids implementation.

For the infinite scroll please note that we ditched some jQuery and removed the fade animation in https://github.com/TryGhost/Casper/commit/41bcbb71570db0f829a2f0f1b5549f380d38d36a because it was killing performance!

flip111:
Removing jQuery code from fitvids.js will be interesting. In my case i can remove jQuery from the theme myself and keep video support.

PR #439, opened 2018-03-06
Fix scrolling of site nav

cotko:
While modifying the casper theme, my site navigation suddenly disappeared when I was scrolling up/down the page.

Example can be found here: https://demo.ghost.io/

casp

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -208,7 +208,6 @@ body {
     overflow-y: hidden;
     -webkit-overflow-scrolling: touch;
     margin-right: 10px;
-    padding-bottom: 80px;
     letter-spacing: 0.4px;
     white-space: nowrap;
 


AileenCGN:
Hey @cotko 

Thanks for the PR. I am not able to reproduce this with Chrome, Firefox, and Safari. Which browser did you have the issue with?

cotko:
@AileenCGN on linux, using chromium, chrome and vivaldi does this, while FF doesn't. Just tried on mac's chrome and safari and it doesn't scroll there using a mouse as well.. It works better using a touchpad which seems to be more precise than mouse wheel.

But if you hover over the *.site-nav-left* element in the dev tools inspector you'll see that the padding makes the element quite big and it seems to be of no use. And if you set the scroll of the element using the dev tools console `$('.site-nav').get(0).scrollTop = 5` then element scrolls everywhere, it just seems that different browsers are differently sensitive to scrolling detection or something:)

JohnONolan:
Padding bottom is required to hide horizontal scrollbars on mobile - this is completely intentional and is explained, in significant detail, in a code comment a couple of lines below your change.

If there's a valid fix which doesn't cause a regression then I'd be happy to merge, otherwise this is a bug which can be reported to Chrome Linux team

PR #441, opened 2018-03-14
Append blog title to post titles to improve SEO

ViViDboarder:
It's often recommended to include the site name in the title of a page for better SEO.

This patch will do so by default for posts and pages.

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -7,7 +7,13 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 
     {{!-- Base Meta --}}
-    <title>{{meta_title}}</title>
+    <title>
+        {{#is "post, page"}}
+        {{meta_title}} - {{@blog.title}}
+        {{else}}
+        {{meta_title}}
+        {{/is}}
+    </title>
     <meta name="HandheldFriendly" content="True" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 


AileenCGN:
Hey @ViViDboarder 

Thanks for the PR! Can you provide some more information, why the site title should be appended to the post title and how it influences the SEO results?

If it turns out to be best practice indeed, I think it would be better to output this information in the existing `{{meta_title}}` helper anyway.

ViViDboarder:
I'm not sure what kind of source you'd like, but here's a blog post explaining a best practice of using the company name (ironically they don't do it themselves) https://www.wix.com/blog/2017/09/how-to-write-seo-title-tag/

Here's some documentation from [Google](https://support.google.com/webmasters/answer/35624?hl=en). There they suggest:

> **Brand your titles**, but concisely. [...] consider including just your site name at the beginning or end of each page title, separated from the rest of the title with a delimiter such as a hyphen, colon, or pipe [...]

Also seen in their title for that page in their "Search Console": *Create good titles and snippets in Search Results - Search Console Help*

Additionally, here's some examples from a few Alexa's top sites:

* [Youtube](https://www.youtube.com/watch?v=dQw4w9WgXcQ)
* [Wikipedia](https://en.wikipedia.org/wiki/Rickrolling)
* [Huffpost](https://www.huffingtonpost.com/entry/rex-tillerson-out_us_59765701e4b0c95f375db326)
* [The Verge](https://www.theverge.com/2018/3/13/17085490/insight-nasa-mars-lander-transport-launch-site-atlas-v)
* [Amazon](https://www.amazon.com/dp/B00MFWQYCO/ref=s9_acsd_bw_wf_a_dlp69917_cdl_1)
* [Twitter](https://twitter.com/vividboarder)
* [Yelp](https://www.yelp.com/biz/vive-la-tarte-san-francisco-2)

I agree inside the `{{meta_title}}` would be a better place. I tried that initially but ended up with this. I can't quite remember why... I can take another look.

AileenCGN:
Thanks! That's a lot of useful information. I will read further into that. But yes, rendering it in `{{meta_title}}` would be the place to put it, IMO. The meta title creation for all SEO purposes happens here: https://github.com/TryGhost/Ghost/blob/master/core/server/data/meta/title.js#L37

AileenCGN:
Nvm what I said about adjusting the `{{meta_title}}`. Adjusting it would mean, that we force it to everyone, which is not desired behaviour, as it's not applicable for some sites.

Adding it to the theme like you did with this PR is a better way, but I would like to hear @JohnONolan's opinion about this change.

JohnONolan:
I'm pretty sure we used to do this and we deliberately stopped doing it because it invalidates the character count of custom meta in the post settings menu. Additionally, on tag archives etc the object title is usually short, like `news` so there is plenty of room to add a site title after the object without going over the character limit. Conversely, the majority of post titles will use the majority of the meta title char limit, not leaving any room for even a relatively short site title.

TLDR: Doesn't make sense for most people based on how ghost is used, and how google treats meta titles. It's a sensible use on a case by case basis, but not a sensible default for everyone.

This is a nuanced subject without a definitive best practice, therefor we stay on the side of doing the thing which fits best for the most people.

For what it's worth, Wix is a terrible source of SEO advice. Here is a significantly better overview of the subject: https://moz.com/learn/seo/title-tag

ViViDboarder:
Moz suggests the same thing. They suggest putting the brand at the end, as in this patch, so that if it does get truncated, your keywords will still be shown.

It seems that it's not actually a fixed width length, so changing to a vertical bar would be a little better.

It would be great if there was a way to pass variables or options into themes so users could decide. As it stands, there's no way to do this without downloading and modifying the theme. 

Anyway, thanks for the review. I figured if I was doing it for myself I might as well see if anyone upstream wanted it. :) Maybe someone else who's trying to figure it out can just pull my branch themselves next time.

PR #443, opened 2018-04-03
Enable CI

mrosenquist:
Using travis ci GSCAN is automatically run

Also, GSCAN is added to lock git hooks for commit and push via the husky lib.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: node_js
+sudo: false
+node_js:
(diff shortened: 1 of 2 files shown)


ErisDS:
This was a really nice idea thanks @mrosenquist! It wouldn't have worked with the old version of GScan because there were no exit codes. I've fixed that, tested it thoroughly and got an updated (slightly simpler) version of this here #470

PR #447, opened 2018-04-09
🐛  Fixed infinitescroll when pagination url used

### LINKED ISSUE #445: Infinite scroll generates incorrect URLs and extra/broken requests
Related to https://github.com/TryGhost/Casper/commit/8690450eaeade583564b0a12c6a2dc12ff729517, infinite scroll still generates additional requests if you're already on page/2 for example

I found this whilst looking at https://forum.ghost.org/t/parameter-page-in-the-url/734

It would be good to do some extra diligence to make sure there is no other way to generate incorrect URLs with the infinite scroll plugin.

### Steps to Reproduce
 
 1. Find some content that has multiple pages
 2. Navigate to the last page e.g. page/2/
 3. Open your network tab to XHR
 4. Scroll down as much as posible
 5. Watch the infinite scroll plugin generate urls like `page/2/page/2/` and try to request them

### Technical details

* Casper Version: version on blog.ghost.org

AileenCGN:
closes #445

- added an fn to sanitize the pathname, that might include a pagination url
- fixed an issue, where the request would still be made if the current page is bigger than max pages
- added comments

### DIFF
--- a/assets/js/infinitescroll.js
+++ b/assets/js/infinitescroll.js
@@ -1,3 +1,5 @@
+/* global maxPages */
+
 // Code snippet inspired by https://github.com/douglasrodrigues5/ghost-blog-infinite-scroll
 $(function ($) {
     var currentPage = 1;
@@ -12,9 +14,6 @@ $(function ($) {
     var lastScrollY = window.scrollY;
     var lastWindowHeight = window.innerHeight;
     var lastDocumentHeight = $document.height();
-
-    // remove hash params from pathname
-    pathname = pathname.replace(/#(.*)$/g, '').replace('/\//g', '/');
 
     function onScroll() {
         lastScrollY = window.scrollY;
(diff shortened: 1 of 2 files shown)


assets/js/infinitescroll.js
```diff
@@ -34,7 +33,29 @@ $(function ($) {
         ticking = true;
     }
 
-    function infiniteScroll() {
+    function sanitizePathname(path) {
+        var paginationRegex = /(?:page\/)(\d)(?:\/)$/i;
+
+        // remove hash params from path
+        path = path.replace(/#(.*)$/g, '').replace('////g', '/');
```
AileenCGN: The original regex for double slash was `/\//g`, which was targeting only a single slash, but according to the commit message [here](https://github.com/TryGhost/Casper/pull/325#issue-129652376) it was supposed to replace double slashes.

assets/js/infinitescroll.js
```diff
@@ -34,7 +33,29 @@ $(function ($) {
...
+
+        // remove pagination from the path and replace the current pages
+        // with the actual requested page. E. g. `/page/3/` indicates that
+        // the user actually requested page 3, so we should request page 4
+        // next, unless it's the last page already.
+        if (path.match(paginationRegex)) {
+            currentPage = parseInt(path.match(paginationRegex)[1]);
```
kevinansfield: Side-effects in single-purpose functions like this are best avoided so that it's easier to follow what's happening.

As an example you could rename this `getPaginationInfo(pathname)` then return an object...

```javascript
var currentPage = 1;

function getPaginationInfo(pathname) {
    ...
    return {
        currentPage: currentPage
        basePath: path
    }
}

function infiniteScroll() {
    var pagination = getPaginationInfo(window.location.pathname);

    // update the outer scope variable to keep track of which page we're on after consecutive loads
    currentPage = pagination.currentPage;

    ...

    var nextPage = pagination.basePath + 'page/' + currentPage + '/';

    ....
}
```

It's still not brilliant but it removes side-effects from the function and paves the way for a later refactor where you have an `InfiniteScroll` object and you can use `this.basePath` and `this.currentPage` rather than mixing global and local variables together.

kevinansfield: Missed it before but this is a good example of what I meant by wrapping the functionality in an object https://github.com/TryGhost/Casper/pull/438/files#diff-c1f39f81b633ffceb749e816cb50c816. With a few changes it would also allow you to pass in the `maxPages` variable so that we're not relying on globals, eg:

```
<script>
    var maxPages = parseInt('{{pagination.pages}}');
    InfiniteScroll.init({maxPages: maxPages});
</script>
```

AileenCGN: Awesome, thanks @kevinansfield! I tried to refactor it, but it took me then to long, so I split it and focused on fixing the bug first. I added this to my todo list for a clean up or bugfix day.

PR #449, opened 2018-04-10
Fix flexbox display in IE

### LINKED ISSUE #376: Formatting in Casper stories in IE11 and Edge.
formatting-on-ghost

### Steps to Reproduce
 
Formatting in stories is all wrong in IE 11.  See image above.  Chrome on right - IE11 on left.
Text and images are the wrong size and all centred in IE11.

### Technical details

* Casper Version: 2.06
* Ghost Version: 1.7.1
* Browser Version: IE11 11.0.10240.17533 (Update Version 110.45)
* OS Version: Win 10

stjosh:
From my point of view, this fixes #376 reliably. Also tested that the layout still behaves correctly with Chrome, Firefox, Safari and Safari on iPhone.


JohnONolan:
This is a pretty obscure set of changes, can you explain how and why this is a valid fix?

stjosh:
Well - the only "real" change I made was in commit 6bd99f1, where I did the following:

- add `align-items: stretch;` to `.kg-card-markdown`: IE11 has quite some issues with flexbox as you can see on [caniuse](https://caniuse.com/#feat=flexbox) or [stackoverflow](https://stackoverflow.com/questions/35111090/text-in-a-flex-container-doesnt-wrap-in-ie11) and other sources. It seems that IE needs this additional hint to render the markdown content correctly and to me, it seemed the least obtrusive way of doing it (idea from [this comment on stackoverflow](https://stackoverflow.com/a/39365207/4361275) - I tried multiple of the suggestions there). Other browsers like Chrome, Firefox and Safari seem to do this by default anyway - at least, rendering of posts was still correct after this change in those browsers.
- set the `max-width: 840px;` to `.post-full-content img` and `.post-full-content video`, as the max-width of their surrounding box is 840px anyway. Otherwise, the images overlap on the right side in IE11 as it seems not to align the images correctly. Also here, this does still look good on the other browsers as far as I can tell.

About the other commits: I just tried to keep my branch mergeable to upstream by pulling some upstream changes. I can also rebase and create a new PR if you'd like.

stjosh:
I agree my changes might be a bit obscure, but I guess IE's flexbox handling is even more obscure. 😄 However, as I want my blog to look good also to users stuck with company-enforced IE11 browsers, I still believe it's worth to merge, especially since it does not break compatibility with other browsers.

kevinansfield:
Hey @stjosh 👋 This PR is in need of a rebase/rebuild 😄

JohnONolan:
Sorry Kevin, I should've replied to this sooner

> set the max-width: 840px; to .post-full-content img and .post-full-content video, as the max-width of their surrounding box is 840px anyway. Otherwise, the images overlap on the right side in IE11 as it seems not to align the images correctly. Also here, this does still look good on the other browsers as far as I can tell.

This is changing the design for all users, not simply fixing a bug for IE11 users - that's why I was asking for more info. Casper's entire post design is predicated on these not being the same widths. So if your fix relies on making them the same width - well then it defeats the object of the code itself. 

So while you're of course very welcome to use these changes on your own fork - atm I can't see that this is a valid fix for core. My feeling is that there is no way around the IE11 issues caused by the current flexbox implementation, and that the real fix will be to do a rewrite of the overall system used to display posts. Possibly with CSS grid.

PR #459, opened 2018-05-31
Use local jQuery instead of CDN

jlelse:
Using the jQuery CDN doesn't comply with GDPR rules afaik. Use a local jQuery version instead of the CDN one.


JohnONolan:
https://forum.ghost.org/t/jquery-cdn-gdpr/1455

anandkumar:
Even if it is fully complaint with GDPR why forcing users to use a third party service? I believe All resources/image/scripts should be loaded locally.

PR #463, opened 2018-07-20
Ghost 1.25 Koenig updates

kevinansfield:
refs https://github.com/TryGhost/Ghost/pull/9741
- Ghost 1.25 has updated the HTML output of Koenig beta posts
  - the `<div class="kg-post">` wrapper around post content has been removed
  - for image cards the `.kg-image-wide` and `.kg-image-full` classes have been changed to `.kg-width-wide` and `.kg-width-full` and applied to the `<figure>` element rather than the `` element
- this will be the default for *all* posts after upgrading to Ghost 2.0

Example output of the `{{content}}` helper for Koenig beta...

Before 1.25.0 (Ghost 1.23.0-1.24.9):
```html
<div class="kg-post">
    <figure class="kg-image-card">
        <img class="kg-image kg-image-wide" src="...">
        <figcaption>example wide image</figcaption>
    </figure>
</div>
```

After 1.25.0:
```html
<figure class="kg-image-card kg-width-wide">
    <img class="kg-image" src="...">
    <figcaption>example wide image</figcaption>
</figure>
```

For reference, in Ghost 1.x a non Koenig post with nothing but an image in the markdown (``) would generate output like this:

```html
<div class="kg-card-markdown">
    <img src="...">
</div>
```


kevinansfield:
@JohnONolan if this looks 👍 then the plan is to merge this, the Ghost-Admin update, and Ghost update with the html/mobiledoc migration together on Monday

kevinansfield:
@JohnONolan there are a number of techniques that work for full-width images without flexbox. For Ghost-Admin's Koenig styles I used variations on the styles in this article https://css-tricks.com/full-width-containers-limited-width-parents/

PR #467, opened 2018-08-02
Using id to query progress

bin16:
Let's query the navbar progress element by id, because some one like me may want to insert progress elements into markdown. By tag name makes them not work.


kevinansfield:
👍 thanks @bin16

PR #470, opened 2018-08-14
🛠 Added .travis.yml and gscan tests

### LINKED ISSUE #105: Sexier typography for posts
Currently the typography on posts could use some love. 
- Headings are not clearly relating to their correct content (closer to the paragraph above or simply sitting in the middle with no clear relation)
- Line-height could use some adjusting, perhaps with a baseline grid to make sure a reading experience isn't tiring

Some helpful posts to better the typography overall beyond just the above:
[Grids from typography](http://blog.weare1910.com/post/72755990718/grids-from-typography?utm_content=buffer1668c&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer)
[Between the lines](http://blog.weare1910.com/)
[Fluid Type](http://trentwalton.com/2012/06/19/fluid-type/)

ErisDS:
Requires us to cut the final 2.0 build of gscan, and this PR to be updated.

- Add gscan dependency & travis.yml file, so that we can run gscan as a test for this repo
- This check tells us whether or not the theme is valid & compatible with Ghost according to gscan

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,9 @@
         "mobile": "assets/screenshot-mobile.jpg"
     },
     "scripts": {
-      "dev": "gulp",
-      "zip": "gulp zip"
+        "dev": "gulp",
+        "zip": "gulp zip",
+        "test": "gscan ."
     },
     "author": {
         "name": "Ghost Foundation",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -41,6 +42,7 @@
     "devDependencies": {
         "autoprefixer": "6.3.6",
         "cssnano": "3.7.1",
+        "gscan": "^2.0.0",
```
ErisDS: @kirrg001 what do you think of pinning gscan like this - so it will install 2.0.1 and even 2.1.0 as we go through the versions? 

Casper should always be up to date :)

kirrg001: 🤘🏽

Issue #466, opened 2018-07-31
Long words do not break causing content overflow


kevinansfield:
### Screenshot

![](https://user-images.githubusercontent.com/4173387/43458425-89cdfdc0-94ca-11e8-9524-fed92944aab8.png)

Sample page: https://jessehouwing.net/visual-studio-2017-merge-tool-in-tower/

### Steps to Reproduce
 
1. Add a very long word with no spaces to a paragraph

### Technical details

* Casper Version: <= 2.4.1

kevinansfield:
Initial testing indicated that we need something like the following applied to children of `.post-content` or `.kg-card-markdown`:

```css
max-width: 100%;
word-wrap: break-word;
```

TonnyORG:
@kevinansfield Fixing this one here: #483

JohnONolan:
^ merged

fadeevab:
Guys, and now it wraps always even if inline code is short:
![Annotation 2019-08-06 191915](https://user-images.githubusercontent.com/5967447/62557292-61a3e880-b87f-11e9-9b9b-5d4d94de0946.png)

JohnONolan:
Nice.

PR #478, opened 2018-09-05
🐛 Fixed word-wrap for extra long anchor links text

AileenCGN:
refs #466

Anchor links didn't break, when the text was too long. Using `word-break: break-all` specifically on a-tags in post-content fixes this without screwing up other tags and their breaks.


peterzimon:
@AileenCGN `word-wrap: break-all` breaks other links also into letters like this (check the "During these retreats" text):

screen shot 2018-09-05 at 11 13 51

I'd recommend using `word-break: break-word;` on `.post-full-content a` instead.

AileenCGN:
Oh well, you got me there... 🙈fixed and pushed

Issue #407, opened 2017-11-08
Links in code blocks are in dark grey on a black background


olivierlambert:
This make those links almost invisible. I'm missing something?

---

### Screenshot

![ghost1](https://user-images.githubusercontent.com/1241401/32556573-1a7594ba-c4a0-11e7-8882-407a5f0d106c.png)

URL live here: https://xen-orchestra.com/blog/docker-support-in-xenserver-the-ultimate-guide/

### Steps to Reproduce
 
 1. latest ghost release
 2. using code blocks AND urls (colored in dark grey)

But visible while hovering:

![ghost2](https://user-images.githubusercontent.com/1241401/32556623-379a82da-c4a0-11e7-9fa6-217c9cdfc5b7.png)

### Technical details

* Casper Version: latest
* Ghost Version: latest
* Browser Version: latest FF and Chrome
* OS Version: Fedora latest

kevinansfield:
@olivierlambert you're not missing anything, looks like you've found a genuine bug, thanks for reporting!

frazjp65:
It is not limited to anchor tags. The same issue can be replicated using "em", "strong", and "h1" tags. I assume there are others too. 

I created a PR (#408) to fix the issue for all descendant elements.

PR #484, opened 2018-10-04
Remove css rule that stops syntax highlighting 

Oshawk:
Removed the ".post-full-content pre code * { color: inherit }" rule which prevents highlight.js from working.

I don't think this breaks anything, please correct me if I am wrong.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -935,10 +935,6 @@ Usage (In Ghost editor):
     background: transparent;
 }
 
-.post-full-content pre code * {
-    color: inherit;
-}
-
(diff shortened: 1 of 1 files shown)


vikaspotluri123:
For reference, these styles were added because of #407

Oshawk:
Hmm, did not realize that. In that case it may be better to make the line less specific by removing the class (although this feels like bad practice).

For reference:
hightlight.js references just by class,
casper references with a class and 2 attributes (`pre code`) making it a little more specific so giving it priority.

Oshawk:
Changed to just exclude span from the rule.

JohnONolan:
v nice approach 👍

Issue #488, opened 2018-10-16
Tag counts are misleading


gargol:
### Screenshots

![image](https://user-images.githubusercontent.com/675397/47032035-966c7500-d171-11e8-9cd0-cbff36c53a82.png)

![image](https://user-images.githubusercontent.com/675397/47032057-ad12cc00-d171-11e8-9080-8bd6cae7647f.png)

### Steps to Reproduce
 
 1. Create a post and tag it with `tag from 2.2.2`. Publish it
 2. Create second post and tag it with same tag: `tag from 2.2.2`. Puslish it (this post ideally has to be created by another user with `author` permissions, but maybe will work for the same author)
 3. Go to the public blog entry of any of the above posts and check the section with tags
 4. The link to posts with current tags says `1 post →` but links to a page with `2` entries

### Technical details

* Casper Version: 
* Ghost Version: 2.2.3
* Browser Version: Google Chrome 69
* OS Version: Ubuntu 18.04

### Code effecting this bug
The [filter](https://github.com/TryGhost/Casper/blob/master/post.hbs#L81) for posts explicitly excludes post that is being displayed:
```
tags:{{primary_tag.slug}}+id:-{{id}}
```
But later on the [wording](https://github.com/TryGhost/Casper/blob/master/post.hbs#L106) explicitly says:
```
plural='See all % posts'
```

kevinansfield:
This could be a tricky one to fix because we can't have nested `{{get}}` helpers. Ideally the `{{primary_tag}}` data that Ghost exposes in the `post` context would include `{{primary_tag.count.posts}}` that way it would be possible to change [the relevant line](https://github.com/TryGhost/Casper/blob/master/post.hbs#L106) to:

```
{{#../primary_tag}}<a href="{{url}}">{{plural count.posts empty='No posts' singular='% post' plural='See all % posts'}} →</a>{{/../primary_tag}}
```

kevinansfield:
A "quick fix" could be to change the wording to indicate more explicitly that it's "x _related_ posts" rather than implying that it's "x total posts"

claudiuconstantin:
„See all x posts in the series” (or something like that) would be a better sentence

PR #489, opened 2018-10-16
🐛 Fix off-by-one error in related tag "all posts" count in post footer

kevinansfield:
closes https://github.com/TryGhost/Casper/issues/488
- re-organises the "related posts" footer column so that we can fetch the primary tag including the `count.posts` value in a separate `{{get}}` block


kevinansfield:
@ErisDS ping in case you know of a better way of doing this 😄 I think a better long-term fix would be for the `{{primary_tag}}` that Ghost exposes to have the `{{primary_tag.count.posts}}` value included

JohnONolan:
I think we agreed this one was overly complex to use RN and we can come back to something better later on

PR #500, opened 2018-11-13
Updated gulp to handle errors for every stream

allouis:
no-issue

Previously we only added an error handler to the first stream in each
pipeline, this meant that if another stream errors you would get
unhelpful error handlers.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,7 +1,7 @@
 var gulp = require('gulp');
+var pump = require('pump');
 
 // gulp plugins and utils
-var gutil = require('gulp-util');
 var livereload = require('gulp-livereload');
 var postcss = require('gulp-postcss');
 var sourcemaps = require('gulp-sourcemaps');
@@ -16,12 +16,6 @@ var cssnano = require('cssnano');
 var customProperties = require('postcss-custom-properties');
 var easyimport = require('postcss-easy-import');
 
-var swallowError = function swallowError(error) {
-    gutil.log(error.toString());
-    gutil.beep();
-    this.emit('end');
-};
-
 var nodemonServerInit = function () {
     livereload.listen(1234);
 };
@@ -32,7 +26,7 @@ gulp.task('build', ['css', 'js'], function (/* cb */) {
 
 gulp.task('generate', ['css', 'js']);
 
-gulp.task('css', function () {
+gulp.task('css', function (done) {
     var processors = [
         easyimport,
         customProperties,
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -1,4 +1,5 @@
 var gulp = require('gulp');
+var pump = require('pump');
```
allouis: This lib takes an array of streams and a "finished" function, pipes the streams together, and calls finishes when the pipeline has finished, or any of the streams error.
```js
pump([streamA, streamB], function (err) {
  // done!
})
```

gulpfile.js
```diff
@@ -17,9 +18,10 @@ var customProperties = require('postcss-custom-properties');
 var easyimport = require('postcss-easy-import');
 
 var swallowError = function swallowError(error) {
-    gutil.log(error.toString());
-    gutil.beep();
-    this.emit('end');
```
allouis: No need to emit end, the streams have finished already

kevinansfield:
@allouis what are we gaining from using the `swallowError` function? From the [`pump` docs](https://github.com/gulpjs/gulp/tree/master/docs/why-use-pump) it seems that gulp's default callback outputs a nice error including the filename and line number, does that still happen with `swallowError`?

`gulp-util` which is used in the `swallowError` message has also [been deprecated](https://www.npmjs.com/package/gulp-util). Maybe we can use the error-handling refactor opportunity to remove that too.

allouis:
@kevinansfield Sorry I missed this! I totally missed the gulp pump docs, I'll fix this up on monday :+1:

kevinansfield:
👍 cool, looks like a good change overall 😄

gulpfile.js
```diff
@@ -41,45 +35,49 @@ gulp.task('css', function () {
         cssnano()
     ];
 
-    return gulp.src('assets/css/*.css')
```
allouis: `return`ing from a gulp task causes the `cb` to be called, which results in errors because we want it to be called once the stream created by pump has ended

allouis:
@kevinansfield This is ready to have a look at again :relaxed:

kevinansfield:
@allouis 👍 looks good 😄 Only issue is the loss of the beep feature which can be pretty useful, maybe we can bring that back in a separate PR?

I don't think pump has an option to specify an error handler so I guess it will be a case of having a `handleError` function, something like this (pseudo-code, not tested)?

```js
var beeper = require('beeper');

function handleError(done) {
    return function(error) {
        if (error) {
            beeper();
        }
        done(error);
    }
}

gulp.task('css', function (done) {
    pump([
        ...
    ], handleError(done));
});
```

JohnONolan:
@allouis could you take a quick look at this ^

allouis:
@JohnONolan On my list to fixup next monday

allouis:
@kevinansfield you happy for this to be merged?

PR #502, opened 2018-11-14
[WIP] Added members drop-in script and dynamic content loader

rishabhgrg:
WIP


post.hbs
```diff
@@ -33,9 +33,29 @@ into the {body} of the default.hbs template --}}
...
-                </div>
+                {{#if @labs.members}}
+                    {{#if html}}
+                        <div class="post-content">
+                            {{content}}
+                        </div>
+                    {{else}}
+                        <div class="post-content"
+                            data-members='true'
+                            data-members-drop-in-resource="post"
```
allouis: Nothing else uses "drop-in" so maybe `data-members-resource-type` might fit better?

post.hbs
```diff
@@ -131,6 +151,11 @@ into the {body} of the default.hbs template --}}
 
 {{!-- The #contentFor helper here will send everything inside it up to the matching #block helper found in default.hbs --}}
 {{#contentFor "scripts"}}
+
+{{#if @labs.members}}
+    <script src="{{asset "js/members-theme-dropin.min.js"}}"></script>
```
allouis: Should this be `built/members-theme-dropin.min.js`?

allouis: I also think this should be in `default.hbs`

post.hbs
rishabhgrg: Yeah thats much better 👍 `drop-in` doesn't sound generic too

allouis: Cool! @rishabhgrg I'm making this change, so can you update the SDK to work with it?

post.hbs
rishabhgrg: Yeah it can go there, it already happens implicitly right now as everything in this block is moved to `defaults.hbs`.

allouis: Yeah, but it'll only happen on a post right, we want the log in/ out buttons on every page afaik :)

rishabhgrg: Ah yes, lets move it 👍

PR #503, opened 2018-11-20
Optimized the implementation of infinite scroll

zce:
no issue

Based on the next page link provided by ghost_head


assets/js/infinitescroll.js
```diff
@@ -1,115 +1,98 @@
...
+        // append contents
+        var postElements = this.response.querySelectorAll('.post-card');
+        postElements.forEach(function (item) {
+            feedElement.appendChild(item);
+        });
 
-        // remove hash params from path
-        path = path.replace(/#(.*)$/g, '').replace('////g', '/');
+        // // push state
+        // window.history.pushState(null, document.title, nextElement.href)
```
kevinansfield: @zce were these comments left in for a reason? I think we probably want to avoid updating the history state in this implementation because it can create unexpected behaviour when sharing URLs

zce: There is no special reason. If you need to hold the state can be achieved in this way

kevinansfield: Can you remove them please? Sorry if my intention for the question wasn't clear 😄 We'd rather not have comments for unused functionality, especially when that functionality isn't directly usable by uncommenting.

zce: No problem

PR #505, opened 2018-11-26
WIP: Responsive images

JohnONolan:


index.hbs
```diff
@@ -29,6 +28,12 @@ into the {body} of the default.hbs template --}}
                 {{!-- The tag below includes the markup for each post - partials/post-card.hbs --}}
                 {{> "post-card"}}
 
+                {{#has number="3"}}
+                    <article class="post-card">
+                        HELLO
```
kevinansfield: :trollface: ?

page.hbs
```diff
@@ -22,7 +22,20 @@ into the {body} of the default.hbs template --}}
...
             {{#if feature_image}}
-            <figure class="post-full-image" style="background-image: url({{feature_image}})">
+            <figure class="post-full-image">
+                {{!-- This is a responsive image, it loads different sizes depending on device
+                https://medium.freecodecamp.org/a-guide-to-responsive-images-with-ready-to-use-templates-c400bd65c433 --}}
+                <img
+                    srcset="{{img_url feature_image size="s"}} 300w,
+                            {{img_url feature_image size="m"}} 600w,
+                            {{img_url feature_image size="l"}} 1000w,
+                            {{img_url feature_image size="xl"}} 2000w"
```
ErisDS: Can I get a hell yeah?

partials/header.hbs
```diff
@@ -0,0 +1,51 @@
...
+
+Ok so, several templates use this big header with a giant BG image. Nice idea, but big images
+have a heavy impact on performance, so it's a good idea to make them responsive. Because we
+can only get the image dynamically using Handlebars, and we can only set the image to properly
+be a background image using CSS, we end up with a handful of inline styles.
+
+If the template in question has a background image, then we render responsive image styles
+for it, and apply those styles to the <header> tag. Else, we just output a <header> tag
+with a `no-image` class so we can style it accordingly.
+--}}
```
ErisDS: Hells Yeah!

package.json
```diff
@@ -57,6 +57,26 @@
         "gulp-uglify": "3.0.1"
     },
     "config": {
-        "posts_per_page": 25
+        "posts_per_page": 25,
+        "image_sizes": {
+            "xxs": {
```
kevinansfield: Is there a use-case for this `xxs` size or is it there more as an example?

index.hbs
JohnONolan: shit

package.json
JohnONolan: actively used for author images @1x and in floating-header.hbs for site icon

kevinansfield: `floating-header.hbs` is using `xs` at the moment, I asked because I couldn't see `xxs` being used anywhere - wondered if it was left in from earlier dev or missed out somewhere else 🙂

post.hbs
```diff
@@ -28,7 +28,20 @@ into the {body} of the default.hbs template --}}
...
+                {{!-- This is a responsive image, it loads different sizes depending on device
+                https://medium.freecodecamp.org/a-guide-to-responsive-images-with-ready-to-use-templates-c400bd65c433 --}}
+                <img
+                    srcset="{{img_url feature_image size="s"}} 300w,
+                            {{img_url feature_image size="m"}} 600w,
+                            {{img_url feature_image size="l"}} 1000w,
+                            {{img_url feature_image size="xl"}} 2000w"
+                    sizes="(max-width: 800px) 400px,
+                            (max-width: 1170px) 700px,
+                            1400px"
```
spaceninja: I know this already got approved and I'm not a team member, but as someone with some responsive image experience, I wanted to ask a few questions. Feel free to let me know if I've missed something obvious.

1) Based on my reading of your `sizes` attribute here, the `300w` asset will never be loaded. You've told the browser the image will be `400px` on small screens, so it'll load the next largest asset, `600w`.

2) I understand you've defined a standard set of image sizes, so they won't always match the exact rendering, but it looks like you used this 400, 700, 1400 scale in several places. Would it make sense to provide assets at 400, 700, and 1400px?

- until 800px, `sizes` says image will be `400px`, browser will load `600w` asset
- until 1170px, `sizes` says image will be `700px`, browser will load `1000w` asset
- above 1171px, `sizes` says image will be `1400px`, browser will load `2000w` asset

Also note that on retina displays, the browser will try to load 2x assets:

- until 800px, `sizes` says image will be `400px`, (800 retina) browser will load `1000w` asset
- until 1170px, `sizes` says image will be `700px`, (1400 retina) browser will load `2000w` asset
- above 1171px, `sizes` says image will be `1400px`, (2400 retina) browser will load `2000w` asset (which is below target width, but the largest asset provided)

JohnONolan: Quite honestly it's my first time implementing responsive images in this format and I've found the whole thing extremely confusing - so if you have an improvement to this PR (and its sounds like you do!) then any help would be extremely welcome 🤗

spaceninja: kk - I'll see what I can do tomorrow :)

spaceninja: Okay, one question before I try to make any code suggestions — is it a problem to add additional sizes to `image_sizes` in package.json? It won't hurt the user, because the browser will just download the appropriate resource, but I'm not sure what your image hosting provider is, or whether generating a few additional sizes for each image would kill your bandwidth costs. (This was a problem at my last job, and it's totally reasonable to strike a compromise between providing the perfect assets vs balancing costs.)

If more assets are fine, then I'd suggest adding 400, 700, 1400 assets to match what you're trying to load on the client.

If you'd prefer to avoid generating additional resources, then I'd look at redefining `s`, `m`, and `l` to 400, 700, and 1400, to better match the most common use cases.

Either way, I'm happy to submit a PR.

Issue #507, opened 2018-12-04
Google Sheets embeds are too small (Integrations)


kymellis:
I am working on the Integrations page for Google Sheets + Ghost, explaining how to embed a spreadsheet into a Ghost post. This is great for people who want to embed dynamic data or swipe files that their readers can export from their site.

### Screenshot

When I use the Google Sheets embed code in a card, it renders very small in Casper:
<img width="885" alt="google sheets embed" src="https://user-images.githubusercontent.com/16795687/49451984-e92de900-f812-11e8-85ef-cd5552267baa.png">

This is not a usable working example of this embed and should fit to the width of the content. In order to publish this integration, it needs to be compatible with Casper.

_Note: the integration content is saved as a draft in the Docs Ghost instance._

kymellis:
@peterzimon found another one of these when moving our podcast embeds from SoundCloud to Backtracks. Live example [here](https://blog.ghost.org/ghost-1-progress/). It should look like [this](https://www.indiehackers.com/podcast/007-john-onolan-of-ghost). 

Some of these integrations are ones I'm working on because people have reached out to us, so if you could let me know when this progresses that would be great :)

kymellis:
@peterzimon scratch the above example - for some reason yesterday pasting the embeddable Backtracks URL into the editor didn't work so I used the embed code in a card instead. It works now and the player renders at the correct size using the OEmbed integration.

It's still an issue for some of the other integrations - but not as much of a blocker for this partnership :)

PR #510, opened 2018-12-15
Set default iframe width and height

peterzimon:
Refs. https://github.com/TryGhost/Casper/issues/507


JohnONolan:
I don't think this approach is viable - there are lots of platforms which use iframes for embeds and we can't guarantee that these dimensions will work for them. I imagine it would break a lot of other embeds

peterzimon:
I'm not sure it's possible to fix this issue with pure CSS (ofc. I'm curious if you have ideas for it). Another approach would be to set a default width and height for iframes that don't have width and/or height attributes via JS? This wouldn't have an effect on either the built in embeds that use iframes nor on the iframes that width/height set.

JohnONolan:
I agree can't be fixed w/ pure CSS - need to come back to this when content flow is non-flexbox or there's a more modern approach

PR #511, opened 2018-12-17
Update text colours to meet WCAG 2.0 AA contrast ratio thresholds

tony-ho:
Addresses the following issue picked up by the [Lighthouse](https://developers.google.com/web/tools/lighthouse/) Accessibility audit:
- Background and foreground colors do not have a sufficient contrast ratio.

See: https://dequeuniversity.com/rules/axe/2.2/color-contrast

The affected colours are blue `#3eb0ef` and mid-grey `#738a94`.

These have been adjusted to `#0f7dba` and `#647a83` as recommended at http://accessible-colors.com/


JohnONolan:
You're very welcome to use these changes locally. The colours used reflect Ghost brand colours and we're not going to change them in core at present.

PR #520, opened 2019-01-10
Use a larger logo size

WesCossick:
Before this change, logos used to display in the size that they were uploaded. Now, the theme uses a 300w logo, which is rather small, especially on high-resolution screens. With Ghost(Pro), you cannot change this setting dynamically, either. You can, however, always upload a smaller logo or downsize a logo with CSS, but you cannot upscale a logo without pixelation. For that reason, I believe it's best to load a larger size.

### DIFF
--- a/index.hbs
+++ b/index.hbs
@@ -7,7 +7,7 @@ into the {body} of the default.hbs template --}}
         <div class="site-header-content">
             <h1 class="site-title">
                 {{#if @site.logo}}
-                    <img class="site-logo" src="{{img_url @site.logo size="s"}}" alt="{{@site.title}}" />
+                    <img class="site-logo" src="{{img_url @site.logo size="l"}}" alt="{{@site.title}}" />
                 {{else}}
                     {{@site.title}}
                 {{/if}}


JohnONolan:
Thanks for the PR! If you want a custom size you can easily use custom CSS with header code injection, or use a custom theme. Both approaches are compatible with Ghost hosted absolutely anywhere, including Ghost(Pro). 

We're not interested in changing the default size at present :)

WesCossick:
Hey @JohnONolan, upscaling a low-resolution image with CSS isn't an option—that just pixelates it. And sure, we could customize the theme and upload it to Ghost(Pro), but that creates a lot of future work to keep it up-to-date as new Casper versions are released every couple weeks. And that route probably isn't an option for non-technical users.

Help me understand your reasoning here. There's little-to-no downside to increasing the default size. Users can _still_ upload a smaller resolution logo if they want, or downscale the logo using CSS (which wouldn't adversely affect the image quality like upscaling does). Why oppose increasing the size?

JohnONolan:
You know what - I actually didn't review this properly and totally misunderstood the change. What you're saying makes complete sense.

JohnONolan:
Apologies for the misunderstanding and thanks for persisting!

PR #550, opened 2019-03-14
Extended `yarn ship`: more automation

kirrg001:
- create draft release on `yarn ship`
- draft release will contain user facing commits between previous and current tag
- keep default release format for now
  - we don't use ghost engine atm, needs further discussion
- full automation with env support (see commit)

Added `.yarnrc` to ensure we keep the tag notation in Casper.
We always used `%s`. `yarn version` forces `v%s` by default.
If we change that suddenly, release ordering breaks.
Ghost+Ghost-Admin use `%s` too.

```
$ yarn ship
yarn run v1.9.4
$ STATUS=$(git status --porcelain); echo $STATUS; if [ -z "$STATUS" ]; then yarn version && git push --follow-tags; else echo "Uncomitted changes found." && exit 1; fi

info Current version: 0.17.0
question New version: 0.18.0
info New version: 0.18.0
Counting objects: 7, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (7/7), done.
Writing objects: 100% (7/7), 763 bytes | 763.00 KiB/s, done.
Total 7 (delta 4), reused 0 (delta 0)
remote: Resolving deltas: 100% (4/4), completed with 3 local objects.
To github.com:kirrg001/casper.git
   5dcb32d..8897d3c  master -> master
 * [new tag]         0.18.0 -> 0.18.0
$ git fetch && gulp release
[12:56:21] Using gulpfile ~/dev/ghost/casper/gulpfile.js
[12:56:21] Starting 'release'...

Draft release for 0.18.0.
Previous version 0.16.2

Release draft generated: https://github.com/TryGhost/Casper/releases/tag/untagged-d65b274f5110a9cb7ce9

[12:56:22] Finished 'release' after 1.51 s
✨  Done in 9.57s.
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,18 +2,18 @@ const {series, watch, src, dest} = require('gulp');
 const pump = require('pump');
 
 // gulp plugins and utils
-var livereload = require('gulp-livereload');
-var postcss = require('gulp-postcss');
-var zip = require('gulp-zip');
-var uglify = require('gulp-uglify');
-var beeper = require('beeper');
+const livereload = require('gulp-livereload');
+const postcss = require('gulp-postcss');
+const zip = require('gulp-zip');
+const uglify = require('gulp-uglify');
+const beeper = require('beeper');
 
 // postcss plugins
-var autoprefixer = require('autoprefixer');
-var colorFunction = require('postcss-color-function');
-var cssnano = require('cssnano');
-var customProperties = require('postcss-custom-properties');
-var easyimport = require('postcss-easy-import');
+const autoprefixer = require('autoprefixer');
+const colorFunction = require('postcss-color-function');
+const cssnano = require('cssnano');
+const customProperties = require('postcss-custom-properties');
+const easyimport = require('postcss-easy-import');
 
 function serve(done) {
     livereload.listen();
@@ -30,7 +30,7 @@ const handleError = (done) => {
 };
 
 function css(done) {
-    var processors = [
+    const processors = [
         easyimport,
         customProperties({preserve: false}),
         colorFunction(),
@@ -56,9 +56,9 @@ function js(done) {
 }
 
 function zipper(done) {
-    var targetDir = 'dist/';
-    var themeName = require('./package.json').name;
-    var filename = themeName + '.zip';
+    const targetDir = 'dist/';
+    const themeName = require('./package.json').name;
+    const filename = themeName + '.zip';
 
     pump([
         src([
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -80,5 +81,8 @@
                 "width": 2000
             }
         }
+    },
+    "dependencies": {
```
kirrg001: dev dep

kirrg001:
The order of releases might be wrong, because `yarn ship` uses e.g. v2.9.0, but casper never used it.
It is a new format. And the last releases created a new tag next to v2.9.0 -> 2.9.0. Hmmmm

kirrg001:
We will stick with `%s` e.g. `2.9.0` tag notation in Casper for now.
All our lerna mono repos use this format. Ghost + Ghost-Admin use this format.

`yarn version` default format is `v%s`. We would need to add a `.yarnrc` file to force this format.

Will discuss for all other repos soon.We have to decide which format we go. Some repos use `v%s` already.

kirrg001:
@rishabhgrg This is ready for review. As soon as we are ready with Casper automation, we will update our internal release doc for Casper 👍

gulpfile.js
```diff
@@ -78,3 +78,118 @@ const dev = series(build, serve, watcher);
...
+const release = () => {
+    // @NOTE: https://yarnpkg.com/lang/en/docs/cli/version/
+    const newVersion = process.env.npm_package_version;
+    let shipsWithGhost = '{version}';
+    let compatibleWithGhost = '2.10.0';
+    const ghostEnvValues = process.env.GHOST || null;
+
+    if (ghostEnvValues) {
+        shipsWithGhost = ghostEnvValues.split(',')[0];
+        compatibleWithGhost = ghostEnvValues.split(',')[1];
```
rishabhgrg: Should we ensure against rogue `ghostEnvValues` here like
```
shipsWithGhost = ghostEnvValues.split(',')[0] || shipsWithGhost;
compatibleWithGhost = ghostEnvValues.split(',')[1] || compatibleWithGhost;
```
So if someone just writes `GHOST=2.x`, `compatibleWithGhost` is not `undefined`. Or maybe thats just too paranoid? 😄

kirrg001: Thought about it, but didn't want to over optimise 🤣

kirrg001: Ok pushed an improvement 👍

rishabhgrg: 😄 Yeah makes sense!

rishabhgrg:
LGTM! 😄 🎉

PR #568, opened 2019-04-22
fixing contrast ratio for "midgrey" text

mansona:
When running a Chrome accessibility audit it said that the text on the "reading time" in a post-card was not high enough. It looks like it gets a score of 3.63: 

Screenshot 2019-04-22 at 09 29 44

This change adds in a `--midgrey-text` variable that should be used in place of `--midgrey` for anything that is on text. The value I chose was literally just dragging the colour picker straight down until it passed the AAA contrast ratio line: 

Screenshot 2019-04-22 at 09 31 46

If you feel like this doesn't live up to the design I think we should at least pass the A test for contrast ratio, but if it's not too far off we should consider passing the AAA test if we can 👍 

Let me know if you have any questions

### DIFF
--- a/assets/css/global.css
+++ b/assets/css/global.css
@@ -10,6 +10,7 @@
     --red: #f05230;
     --darkgrey: #15171A;
     --midgrey: #738a94;
+    --midgrey-text: #4a5b62;
     --lightgrey: #c5d2d9;
     --whitegrey: #e5eff5;
     --pink: #fa3a57;
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -432,7 +432,7 @@ The knock-on effect of this is ugly browser-scroll bars at the bottom, so 80px o
 .post-card-tags {
     display: block;
     margin-bottom: 4px;
-    color: var(--midgrey);
+    color: var(--midgrey-text);
     font-size: 1.2rem;
     line-height: 1.15em;
     font-weight: 500;
@@ -585,7 +585,7 @@ The knock-on effect of this is ugly browser-scroll bars at the bottom, so 80px o
 .reading-time {
     flex-shrink: 0;
     margin-left: 20px;
-    color: var(--midgrey);
+    color: var(--midgrey-text);
     font-size: 1.2rem;
     line-height: 33px;
     font-weight: 500;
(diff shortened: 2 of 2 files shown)


mansona:
If you want to see what this looks like to evaluate the colour change you can check it out here: https://deploy-preview-20--empress-blog-casper-template.netlify.com/ 👍

mansona:
Any thoughts on this? If you would prefer that I change this to only pass AA rating then let me know and I'll update the PR 👍 that would have less impact on the design 😄

gargol:
Hey @mansona :wave: I think these are similar changes to ones addressed here - https://github.com/TryGhost/Casper/pull/511#issuecomment-447826877. Thanks for the contribution anyway :+1:

mansona:
@gargol well I'm not 100% sure they are the same changes 🤔 I added a new variable `--midgrey-text` and used that only for text that previously used `--midgrey`

I understand what @JohnONolan said about not wanting to change brand colours, and that's why I made sure to only change the text that was affected. 

Would you be more likely to accept a PR if I only made it pass the AA test instead of AAA?

JohnONolan:
Hi Mansona - the reason Ghost is open source is so that you can make changes like this - however you want to. You aren't limited by the product we offer, you can make it your own based on what is most important to you.

You're very welcome to use your own colours in your own version of Casper - there's no problem at all with that.

mansona:
@JohnONolan yes I accept that 😄 and I know that I can make a change in the downstream project without too much difficulty.

What I'm more interested in overall is having as few as possible custom changes in our fork for two reasons: a) it's easier to maintain and b) if we're making a change to improve the accessibility of the Casper template I would personally much prefer that change to be available to the most people possible.

My last question was really about what it would take to get this (relatively small) accessibility change upstream. If the Casper template is not planning to conform to [WCAG](https://www.w3.org/TR/WCAG20/) level AA then that's fine, and we can make whatever changes we want downstream 👍 I am just trying to open a dialogue to try to get it upstreamed if at all possible 🎉

JohnONolan:
You can easily add a colour change as an inline style via the code injection feature - which requires no fork or maintenance of any kind, so that should be straightforward enough

As mentioned, we don't have any plans to change the upstream colours at present. If/when we do make changes to style or design - many different factors are brought into consideration and we go through that process with the product and design teams, considering impact across our wider design system, projects, and several hundred repositories. We don't make changes like this on an ad-hoc basis

PR #580, opened 2019-05-11
updated Facebook icon to match new branding

tomhmoses:
- new icon from: https://en.facebookbrand.com/assets/f-logo/
- converted to svg using: https://www.zamzar.com/
- tested on https://www.ubmc.co.uk/


JohnONolan:
Thanks for this! We've update the icon now with a compressed version of this SVG and will release soon :)

PR #581, opened 2019-05-12
Update jQuery cdn

0xsuid:
Updated jQuery source url from code.jquery.com to cdnjs for better loading speed of websites.

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -69,7 +69,7 @@
 
     {{!-- jQuery + Fitvids, which makes all video embeds responsive --}}
     <script
-        src="https://code.jquery.com/jquery-3.2.1.min.js"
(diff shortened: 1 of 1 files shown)


0xsuid:
@allouis @kevinansfield @peterzimon @JohnONolan @ErisDS :ping_pong: 

Hope This ping represents 4-5 PRs that is pending from few days and will not ban my PR/comment regarding code of conduct.

JohnONolan:
I'm not sure what that means but we don't have any intention of changing jQuery source at the moment. We're very deliberately using the foundation's CDN.

0xsuid:
cdnjs is better cdn then Foundation's cdn, you can use gtmetrix or pingdom to find out that.

cdnjs helps to reduce loading time of it by providing better cdn which is provided by cloudflare.

PR #584, opened 2019-05-23
Fix Safari render/paint loop

happycollision:
Fix Safari render/paint loop

**What:**

As identified in [this bug
report](https://bugs.webkit.org/show_bug.cgi?id=194332) with additional
comments clarifying my findings hopefully soon to appear, there is a bug
in Safari that starts and endless cycle of render/paints when all the
following conditions are met:

1. a `transition` animation is defined that is not instantaneous
2. In the destination of the transition, an animating property contains
   a color value with:
  a) an alpha channel that is *exclusively* between 0 and 1
  b) at least one color channel which is *exclusively* between 0 and 255

These conditions represent my current best guess at the exact
requirements to encounter this bug.

You can explore the bug in this fiddle I made while trying to pinpoint
the minimum reproduction: https://jsfiddle.net/0ph9xkrt/7/

The bug described above affects this template only on the `.post-card`
class. All other animations (that I am aware of and have tested)
involving an element with an `rgba` color in one of its properties do
not directly animate the property containing the color, and are thus not
triggering this bug.

**How:**

There are three ways we can avoid the bug for the hover animation of
`.post-card`s in this template. One, we could just alter the shadow's
color to be a pure black (nullifying condition 2.b) and then compensate
the darker color with a slightly more transparent alpha channel. Two, we
could use the actual design colors (which actually contain a hint of
color) but skip animating the distance/blur of the shadow which means we
do not have to animate a property containing the color (nullifying
condition 2). Or three, we can use pseudo-selectors to define `:before`
and `:after` with each representing one state for the box shadow, then
animate only the opacity between the two (also nullifying condition 2).

This third option was the choice here since it will keep the design of
the site as is, visually, without compromising the color of the shadow
or the changable size of the shadow. It does have one drawback that the
other two options do not: the css is much more complex, as you'll see in
the diff. Not only do we need to add all the different states for two
new pseudo-selectors, but we also need to ensure that the links under
them are set to have a higher z-index so they remain clickable.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -387,20 +387,51 @@ The knock-on effect of this is ugly browser-scroll bars at the bottom, so 80px o
     flex: 1 1 300px;
     display: flex;
     flex-direction: column;
-    overflow: hidden;
+    overflow: visible;
     margin: 0 20px 40px;
     min-height: 300px;
     background: #fff center center;
     background-size: cover;
     border-radius: 5px;
-    box-shadow: rgba(39,44,49,0.06) 8px 14px 38px, rgba(39, 44, 49, 0.03) 1px 3px 8px;
     transition: all 0.5s ease;
+    z-index: 1;
+    position: relative;
 }
 
 .post-card:hover {
-    box-shadow: rgba(39,44,49,0.07) 8px 28px 50px, rgba(39, 44, 49, 0.04) 1px 6px 12px;
     transition: all 0.4s ease;
     transform: translate3D(0, -1px, 0) scale(1.02);
+}
+
+.post-card:before, .post-card:after {
+    display: block;
+    content: '';
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    z-index: 0;
+    transition: all 0.4s ease;
+    border-radius: 5px;
+}
+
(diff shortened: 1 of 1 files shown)


kevinansfield:
Hey @happycollision 👋 Thanks for looking into this and providing such a detailed investigation into the bug!

The changes in this PR have been reviewed internally and it was decided the third option increases the CSS complexity a bit too much - Casper is meant to be as beginner-friendly as possible. I've got some time set aside for next Monday to investigate the alternative options you've laid out to see if we can find a good balance between CSS complexity and design impact, I'll keep you posted 🙂

happycollision:
Makes sense, for sure. That is exactly why I pointed out the added complexity.

I believe that my first suggested alternative will keep you most closely in line with what the current design is. I am guessing that such a subtle change in color (brighter color, lower transparency -> darker color, higher transparency) is less noticeable than a change in behavior (animating shadow spread -> no animation on the shadow). If your shadow happened to be further away from a strict grey than it is, I'd say both alternatives would be quite noticeable.

Of course, you could always leave it as is, and count on the bug to be fixed in Webkit. Though, who knows when that will be...

JohnONolan:
I believe this issue is redundant in V3, which removes these shadows - so it should be fixed now by proxy :)

PR #599, opened 2019-07-30
Better contrast for selected text inside `pre code` blocks

mvasilkov:
Hi, I've encountered an issue where selected text inside code blocks is barely visible. To fix it, I changed the color of the `::selection` pseudo to match the color of body copy.

Before:
Before: low contrast selected text

After:
After: normal contrast selected text


daviddarnes:
@mvasilkov looks great, thanks for the fix!

PR #602, opened 2019-08-01
Prevent code cards with long lines from being too wide

mvasilkov:
Another small issue that I found: when we have a code card with caption, on a small screen it can become wider than the viewport. A code card without caption receives a horizontal scroll bar, which I think is the correct behavior. This is exemplified in the following screenshot:

Screen Shot 2019-08-01 at 17 03 35

Without caption, a code card becomes a `<pre>` tag with `max-width: 100%`. With caption, a code card renders as a `<figure class=kg-code-card>` tag with `min-width: 100%`. The lack of max-width allows the figure's contents to overflow the viewport.

To alleviate this, I merged min- and max-width into just `width: 100%` on the `kg-code-card` class. Now code cards behave the same, with or without caption:

Screen Shot 2019-08-01 at 17 04 50


rishabhgrg:
Closed in https://github.com/TryGhost/Casper/commit/8bb4508ff2ed99bc7de4414c4bc491eaed0dd179. Thanks @mvasilkov for the fix :)

PR #609, opened 2019-09-03
Fixes background color issue with post-card image rounded corner

swelham:
This PR fixes an issue I noticed where there is a small amount of white background being rendered on the top left `.post-card` rounded corner.

For example (you might need to zoom in a little).

Screenshot 2019-09-03 at 19 43 34

Moving the background-color onto the `. post-card-content` class solves this issue.


JohnONolan:
Happy to investigate this if your can provide a working copy of the problem - atm can't verify what's happening or whether this is the right solution, based on the provided screenshot

PR #612, opened 2019-09-19
Changed {{code}} to {{statusCode}}

gargol:
refs https://github.com/TryGhost/gscan/commit/2ebd9feeee20c084a1d4ab84e5100cdb1e1400de

- {{code}} use has been deprecated in canary rule set of gscan

TODO:
- [x] do a real world test with theme with these changes


gargol:
The CI is erroring becuse `gscan` is being run with v2 set of rules instead of `canary`

PR #615, opened 2019-10-02
Casper 3.0

peterzimon:
Refreshed the overall theme. High level summary of the changes:
*Site level changes*
- refined and simplified site header
- refined mobile sizes

*Post lists*
- redesigned post cards
- added adaptive site headers for tag and author lists depending whether they have background image

*Posts*
- refined typography and vertical rhythm
- removed read indicator
- unified author avatars for cards and posts
- simplified feature image
- added custom post excerpt
- redesigned read next footer


partials/post-card.hbs
```diff
@@ -22,40 +22,46 @@
...
                     {{/if}}
                 </li>
-            {{/foreach}}
+                {{/foreach}}
             </ul>
-
-            <span class="reading-time">{{reading_time}}</span>
-
+            <div class="post-card-byline-content">
+                <p>{{#has author="count:>2"}}Multiple authors{{else}}{{authors}}{{/has}}</p>
```
peterzimon: to be checked: when there are more than 2 authors we display "Multiple authors"

assets/css/screen.css
```diff
@@ -152,90 +153,198 @@ body {
...
     min-height: 200px;
-    max-height: 450px;
-    text-align: center;
+    max-height: 340px;
 }
 
 .site-title {
     z-index: 10;
-    margin: 0;
+    margin: 0 0 0 -2px;
```
peterzimon: to be checked: bigger size SF fonts is optically incorrect and looks like it's vertically misaligned with the smaller sized text. not sure this is needed as SF is only default for MacOS/iOS

assets/css/screen.css
```diff
@@ -152,90 +153,198 @@ body {
...
     font-weight: 300;
-    letter-spacing: 0.5px;
-    opacity: 0.8;
+}
+
+
+/* 4.2 Archive header (tag and author post lists)
+/* ---------------------------------------------------------- */
+
+.site-archive-header .site-header-content {
```
peterzimon: to be checked: `.site-archive-header` refers to tag and author post lists, `.site-home-header` to home (design and behaviors are different). unsure about naming and if we have to differentiate at all, i.e. we could just have `.site-header` for archives and differentiate `.site-home-header` or vice versa

assets/css/screen.css
```diff
@@ -375,39 +570,36 @@ The knock-on effect of this is ugly browser-scroll bars at the bottom, so 80px o
 /* 6. Post Feed
 /* ---------------------------------------------------------- */
 
+.posts {
```
peterzimon: to be checked: each post card has a lightgrey bottom border (and they don't have any margins) in order to have vertical separation between the rows in the post list. the card list layout depends on many things (whether the post has image, the number of posts etc.) and I couldn't find a definite CSS rule to solve this in another way. 
this is a hack to hide the left and right side of these separator lines. unsure if this is okay / if there's any other way to solve it.

assets/css/screen.css
```diff
@@ -1162,16 +1457,198 @@ Usage (In Ghost editor):
...
+    }
+
+    .post-full-byline-meta h4 {
+        font-size: 1.2rem;
+        margin-bottom: 2px;
+    }
+}
+
+
+/* 7.2. Subscribe Form
```
peterzimon: to be checked: do we keep the subscribe form? we'll replace subscriptions with members in 3.0, though this theme can be still used with 2.x installs...

partials/post-card.hbs
JohnONolan: I'm trying to figure out how this is used but not sure I understand this - can you provide context via screenshots maybe?

assets/css/screen.css
JohnONolan: I think this is ok

assets/css/screen.css
JohnONolan: sounds like a good one for @daviddarnes but might be helpful if you provide more context with screenshots

assets/css/screen.css
JohnONolan: TBC this week

JohnONolan:
NB: This has been merged to the v3 branch here: https://github.com/TryGhost/Casper/pull/617 -- all further work should be done there, then when we're ready to ship we merge the v3 branch to master

JohnONolan:
Moving here: https://github.com/TryGhost/Casper/pull/618

PR #619, opened 2019-10-16
V3 darkmode

peterzimon:
- added OS dark mode support


assets/css/global.css
```diff
@@ -14,6 +14,7 @@
     --whitegrey: #e5eff5;
     --pink: #fa3a57;
     --brown: #a3821a;
+    --dark-theme-background: color(var(--darkgrey) l(+2%));
```
peterzimon: @JohnONolan I'm not sure this is the correct place and naming for this CSS var

PR #679, opened 2020-02-05
Fixing image bug for link preview

Amoenus:
This PR is to fix the default behaviour of image previews for default Casper theme.

After the change this:
image
Becomes this:
image

---
This change is [Reviewable](https://reviewable.io/reviews/tryghost/casper/679)


JohnONolan:
You're welcome to make this change locally if you prefer it - but this is stylistic choice, not a bug :)

PR #685, opened 2020-02-19
Clarify purpose of dev server in README

miohtama:
As it is not clear from the README how the theme developers should go around and start working if you land here from an unknown source

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,10 @@ yarn install
 yarn dev
 ```
 
+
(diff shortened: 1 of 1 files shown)


daviddarnes:
Hey @miohtama, appreciate the contribution but we rarely ever encounter people landing on a theme without knowing about local theme development. We'd also prefer to keep these types of documentation in one central location to prevent inconsistencies 🙂

PR #701, opened 2020-04-01
Post cards > Lazyload

rafaelstz:
I added the native lazy load in the post images.
Reference about it: https://web.dev/native-lazy-loading/


JohnONolan:
Good idea, thanks!

PR #705, opened 2020-04-16
Change border-left on blockquote to use color variables

icaroharry:
The blockquote is using a hardcoded color in `screen.css`. I've changed it to use the `color(var(--primary))` instead.


JohnONolan:
Thanks!

Issue #713, opened 2020-05-20
gulp-zip sets permissions incorrectly on Windows


moritonal:
Howdy, it seems that the gulp process for Casper on Windows zips up permissions in a way Ghost (inside docker) doesn't like due to a bug in `gulp-zip`.

It seems this is caused by an issue in [gulp-zip](https://github.com/sindresorhus/gulp-zip/issues/76). There are some PR's trying to fix it but they are blocked by issues lower down in the dependency tree.

### Screenshot

Uploading the new ZIP of a custom casper fires off permission issues creating a directory in `assets`.

![broke_upload](https://user-images.githubusercontent.com/6094389/82443035-b1fbd800-9a98-11ea-9bbf-198920b3749b.PNG)

### Steps to Reproduce
 
 1. Download Casper Theme locally
 2. Change something
 3. Run `npm run zip`

### Technical details

* Casper Version: 3.0.12
* Ghost Version: 3.16.0
* Browser Version: Firefox (evergreen)
* OS Version: Windows 10

### Solution

Seems that simply unzipping then re-zipping Casper cheers it up. I've attached a PR which swaps over to the `gulp-vinyl-zip` package for zipping which seems to work fine for me if you'd like to use it.

On a side note, has there never been a Windows dev for a Ghost theme!?

moritonal:
This seems to show the main issue here. Left is the permissions without the fix, right is with the fix.

![image](https://user-images.githubusercontent.com/6094389/82448924-520a2f00-9aa2-11ea-87f9-937e68ae2e76.png)

ErisDS:
This was fixed upstream, and should be working as expected now.

PR #714, opened 2020-05-20
Fix issue 713

moritonal:
_Fix_ for issue #713, although really it's just shuffling over the dependency you use for zipping. I'll compare a zip produced by before and after and see that changes.


ErisDS:
Thanks for the patch @moritonal! 

Going to close however as this was fixed upsteam in gulp-zip

PR #716, opened 2020-05-25
Translations

TheM1984:
Confirm the guide on [translations](https://ghost.org/docs/api/v3/handlebars-themes/helpers/translate/) we made the Casper theme translatable.

This way the theme could easily be used in a multilingual setup.

Maybe more translations could be added in the future, or could be added manually by users.


JohnONolan:
You're very welcome to translate the theme or release your own translated themes! We're not ready to include translations by default in Casper at the moment

PR #733, opened 2020-08-02
Fix background color of slash of darkmode <hr>

stromfeldt:
I submitted [the Lyra fix](https://github.com/TryGhost/Lyra/pull/16) for this issue in late-2019, and having just noticed it also exists in Casper, here's the same fix.


JohnONolan:
Thanks again!

PR #734, opened 2020-08-29
Dismiss message automatically

groninge01:
Just dismiss the subscribe success message after 5 seconds without any user interaction.


JohnONolan:
Thanks for the suggestion! We'll be overhauling these notifications shortly in a slightly different way :)

PR #739, opened 2020-10-01
Translation Support

JamilFilho:
Casper is the default Ghost theme and in many cases is not changed by the user. So it would be interesting to add support for translations.

Following the Documentation guidelines, two files were created in `content/themes/mytheme/partials` to allow Casper to be fully translatable:

**pagination.hbs**
**navigation.hbs**

In `locales` have already created a Portuguese translation (`pt.json`), which can be used as a basis for translating Casper into any other language.


vikaspotluri123:
Closing as a duplicate of #716

PR #764, opened 2021-03-11
Only use post-card-large variation if feature_image exists

PaulAdamDavis:
If the first post has no `feature_image`, it was still incorrectly displayed as a large card.

This PR adds another conditional to check `feature_image` exists. With this, the large variation will only be used if:
* is home page
* `feature_image` exists
* is the first newest post

### DIFF
--- a/partials/post-card.hbs
+++ b/partials/post-card.hbs
@@ -1,7 +1,7 @@
 {{!-- This is a partial file used to generate a post "card"
 which templates loop over to generate a list of posts. --}}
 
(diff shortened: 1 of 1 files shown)


JohnONolan:
Does this break tag archives? They depend on this code too and look (intentionally)  more like the "before" screenshot, in that context

PaulAdamDavis:
The tag archive has `<header class="post-card post-card-large">` hard-coded into it, so is unaffected.  This only affects posts within the loop.

Here's a screenshot of this PR in action, on a tag archive:

Screenshot 2021-03-11 at 18 38 39

JohnONolan:
aaah, awesome

PR #765, opened 2021-03-11
Fix horizontal scrolling caused by scaling on the subscribe CTA h2

PaulAdamDavis:
`transform: scale(1.15);` on the subscribe CTA `h2` caused horizontal scrolling on smaller screens, as illustrated by the red outlines in the before screenshot.

This PR removes that `transform` in favour if increasing the font-size by 15%, matching the same visual size that the transform produced.

## Before:

Screenshot 2021-03-11 at 17 01 37

## After:

Screenshot 2021-03-11 at 17 08 17


JohnONolan:
Agree with approach, but if we're going to manually specify font sizes then lets just specify absolute values in rem rather than calc * 1.15 - makes it a bit more readable

PaulAdamDavis:
Absolute valued was my first approach, but I thought this would make find/replace easier. Agreed though, will change.

JohnONolan:
@PaulAdamDavis Media query style now specifies EM rather than REM, is that intentional or typo?

PaulAdamDavis:
@JohnONolan Typo, thanks for spotting! Fixed.

PR #768, opened 2021-03-12
Fix multiple author layout

minimaluminium:
This PR fixes post card and single post byline layout when there are many authors (more than 3).

### Post card (before)

Screen Shot 2021-03-12 at 11 21 28 AM

### Post card (after)

Screen Shot 2021-03-12 at 11 15 12 AM

### Single post byline (before)

Screen Shot 2021-03-12 at 11 20 52 AM

### Single post byline (after)

Screen Shot 2021-03-12 at 11 15 30 AM


JohnONolan:
@minimaluminium Don't want to add additional markup for handling an edge case here - it makes the HBS file harder to read/understand for everyone, when it will be used by almost no-one. Casper needs to stay as minimal as possible in code so that it's easy for beginners to understand - so I would rather limit this to a single max-width and let it be imperfect

JohnONolan (changes requested):
@minimaluminium Don't want to add additional markup for handling an edge case here - it makes the HBS file harder to read/understand for everyone, when it will be used by almost no-one. Casper needs to stay as minimal as possible in code so that it's easy for beginners to understand - so I would rather limit this to a single max-width and let it be imperfect

minimaluminium:
Agree, we shouldn't complicate the HBS templates for edge cases like this. Reverted it, and set `max-width` on author list.

PR #772, opened 2021-03-12
Increase spacing between a heading and anything after it

minimaluminium:
### Before

Screen Shot 2021-03-12 at 12 32 58 PM

### After

Screen Shot 2021-03-12 at 12 32 44 PM


JohnONolan:
This is intentional atm

PR #778, opened 2021-03-18
hide code block scroll bar

heng1025:


JohnONolan (changes requested):
Please provide screenshots, details, and a reason for the change. I’m interested but we don’t generally merge PRs without any context

heng1025:
version: 4.01,4.1.0

before:
image

after:
image

PR #817, opened 2021-09-22
patch: adjust Casper theme

rafalbromirski:
### What

Adjust casper theme:
- remove native "subscribe" (button + footer box)
- add secondary navigation (a bit broken on smaller screen but we will get there)
- dark mode icon + basic script to handle it


rafalbromirski:
sorry.. wrong place 😮‍💨

PR #819, opened 2021-10-15
Footer > Adding Lazy loading in the profile image

rafaelstz:
I'm adding lazy loading in the footer profile image as we have to the post card image as it will be out of the browser viewport on mobile.


minimaluminium:
Hey @rafaelstz, looks great! Thank you 🙂

PR #822, opened 2021-10-17
Make <kbd> tag look more like a button

artemkovalyov:
The kbd tag is often used in tech articles and it doesn't look nice at the moment. By the way, it's also looking wrong in the editor.
That's how it should look after this change.

image

### DIFF
--- a/assets/css/global.css
+++ b/assets/css/global.css
@@ -169,8 +169,21 @@ svg:not(:root) {
 mark {
     background-color: #fdffb6;
 }
+kbd {
+  display: inline-block;
+  margin: 0, 0.1rem;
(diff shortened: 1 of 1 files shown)


minimaluminium:
Hey @artemkovalyov, thanks for your PR! I've just closed this one, and added `<kbd>` support in https://github.com/TryGhost/Casper/commit/1f3418708c869e819e5f565aad91ada2765f13e1. This is how it looks in light and dark mode.

CleanShot 2022-01-03 at 12 07 15@2x
CleanShot 2022-01-03 at 12 07 27@2x

artemkovalyov:
Amazing, @minimaluminium, thanks for investing into making it look like it should:)

PR #870, opened 2022-06-08
Add `me` relation to social links

jeromegamez:
Hi there 👋,

The [Microformats spec](https://microformats.org/wiki/rel-me#example) says that

> rel="me" is used on hyperlinks from one page about a person to other pages about that same person. 

It's no hard argument, but at least Github adds this to URLs as well 😅, and [Twitter recommends it in their developer docs](https://developer.twitter.com/en/docs/twitter-for-websites/webpage-properties) (in the "Identify the Twitter profile of the page" section).

:octocat:

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -44,10 +44,10 @@
             <div class="gh-head-actions">
                 <div class="gh-social">
                     {{#if @site.facebook}}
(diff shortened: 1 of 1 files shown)


minimaluminium:
Thanks @jeromegamez, the links were helpful as well 🙂

PR #874, opened 2022-06-16
Fix author profile location is dark color in dark mode

ViaSnake:
When the color scheme is "Dark" or "Auto" and the device is in dark mode, the location text on the author page is displayed in white.

---

## Before
image
image

## After
image


minimaluminium:
Hey @ViaSnake, nice one! Thanks for your contribution.

PR #876, opened 2022-07-07
Fix missing space between paragraphs within a blockquote.

markstos:
I'll upload screenshots to show the before/after difference.

## Before

No space between paragraphs in blockquotes.
image

## After

Space between paragraphs in block quotes,  but not at the top or bottom. The blockquote already provides margin around the whole feature.

image

If users have multiple paragraphs within a blockquote, presumably there are there for a reason. If you don't want the space in your blockquotes, don't use multiple paragraphs.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -1081,6 +1081,18 @@ is the very first element in the post content */
     margin-top: max(4.8vmin, 32px) !important;
 }
 
+/* Within a blockquote, no margin is needed at the top or bottom
+but do include space between multiple paragraphs */
+.gh-content > blockquote p:first-child {
+    margin-top: 0;
+}
+.gh-content > blockquote p {
+    margin: 1.5rem;
+}
+.gh-content > blockquote p:last-child {
+    margin-bottom: 0;
+}
+
 /* Now the content typography styles */
 .gh-content a {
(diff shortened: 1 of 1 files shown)


minimaluminium:
Hey @markstos,

Blockquotes don't contain paragraphs when added from the editor. If space is needed like shown on your screenshot, it should be added by pressing <kbd>Shift+Enter</kbd> like below.

The editor:
CleanShot 2022-10-04 at 2 34 44@2x

The output:

CleanShot 2022-10-04 at 2 38 10@2x

I assume your fix applies to blockquotes added by a HTML card. We shouldn't be complicating the theme CSS for such an edge case.

markstos:
Thanks for the reply @minimaluminium. I've read a lot of Ghost docs, but I missed the tip about "Shift+Enter". Where is that documented? It seems usual to mention at "Using the Editor" https://ghost.org/help/using-the-editor/

minimaluminium:
The combination of <kbd>Shift+Enter</kbd> is called soft-break and it adds a line break instead of creating a new paragraph. It's also used widely on text editing applications.

When typing in a blockquote, it exits the blockquote and jumps to a new paragraph when you press <kbd>Enter</kbd>. However, it won't exit the blockquote when you use the soft-break :)

Issue #887, opened 2022-07-26
GulpUglifyError: unable to minify JavaScript


gllmhyt:
Here's the error:
```
yarn run v1.22.19
warning casper@5.2.2: The engine "ghost" appears to be invalid.
$ gulp zip
[06:30:55] Using gulpfile [...]/Casper/gulpfile.js
[06:30:55] Starting 'zip'...
[06:30:55] Starting 'css'...
[06:30:56] Finished 'css' after 728 ms
[06:30:56] Starting 'js'...
[06:30:56] 'js' errored after 140 ms
[06:30:56] GulpUglifyError: unable to minify JavaScript
Caused by: SyntaxError: Unexpected token: name «s», expected: punc «;»
File: [...]/Casper/assets/js/lib/casper.js
Line: 208
Col: 1113
[06:30:56] 'zip' errored after 870 ms
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
May I consider switching to [terser](https://github.com/terser/terser)?

minimaluminium:
Hey @gllmhyt, the error is in `/assets/js/lib/casper.js` which doesn't exist. Is this a custom JS file?

gllmhyt:
@minimaluminium That's really weird, I have no such file...

minimaluminium:
I couldn't reproduce the issue on my end. Could you try pulling the latest changes, delete `/node_modules` and run `yarn` and `yarn dev` again? I assume you got this error while trying to run `yarn dev`?

gllmhyt:
Indeed when I clone the upstream branch I don't have that problem, I will try to rebase my branch against it and see what's causing the problem

minimaluminium:
Sounds great! We're using gulp-uglify for all our themes, and if we need to replace this, we'd also need to replace it everywhere. We're not planning to change this at the moment. We really appreciate your idea and contribution 🙂

PR #888, opened 2022-07-26
Use gulp-terser instead of gulp-uglify

gllmhyt:


minimaluminium:
For the reason https://github.com/TryGhost/Casper/issues/887#issuecomment-1196265912

PR #908, opened 2022-11-12
Fix intended location of search icon

tmlmt:
The match helper was not referring to the right custom setting


minimaluminium:
That's a great one! Thanks @tmlmt :)

Issue #907, opened 2022-11-11
Social icons with 5.4.0


viriatusX:
Hi,
with the latest Casper v.5.4.0 the option to add social icons to header has dissapeared?
It's very useful....
Can it be implemented again or explain how to recover this function please?
Thanks!

minimaluminium:
Hey @viriatusX, the removal of social icons in `v5.4.0` was intentional which is a part of standardizing headers across all our official themes. Our current Facebook and Twitter links are structured data for SEO purposes, and won't be used on our themes anymore.

If anyone wants to bring back the links to their websites, the code can be taken from `v5.3.2` [here](https://github.com/TryGhost/Casper/blob/c7713df88f11bc2ff612457dfb51d6e1909a40e4/default.hbs#L49-L55).

minimaluminium:
Just FYI, we now have a dedicated tutorial for adding social icons to the official themes with code injection method. https://ghost.org/tutorials/add-social-media-icons/#code-generator

viriatusX:
> Just FYI, we now have a dedicated tutorial for adding social icons to the official themes with code injection method. https://ghost.org/tutorials/add-social-media-icons/

Thanks @minimaluminium! but it doesn't work with Casper 5.4.0. I just tried it now...

PR #909, opened 2022-11-12
Fix disappearance of social media icons (#907)

tmlmt:
Proposal to restore social media icons which were removed by #906 in v5.4.0
Independent from my other very simple pull request, as the maintainers may prefer different css/layout choices that those made here.

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -430,6 +430,7 @@ production stylesheet in assets/built/screen.css
 
 .gh-head-menu {
     display: flex;
+    gap: 32px;
     align-items: center;
     margin-top: 1px;
     font-weight: 500;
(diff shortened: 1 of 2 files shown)


minimaluminium:
Hey @tmlmt, thanks for the PR! However, the removal of social icons in v5.4.0 was intentional which is a part of standardizing headers across all our official themes. Our current Facebook and Twitter links are structured data for SEO purposes, and won't be used on our themes anymore.

If anyone wants to bring back the links to their websites, the code can be taken from v5.3.2 [here](https://github.com/TryGhost/Casper/blob/c7713df88f11bc2ff612457dfb51d6e1909a40e4/default.hbs#L49-L55).

PR #916, opened 2022-12-22
Fix  match the @custom.navigation_layout in default.hbs

officialrajdeepsingh:
Fix the issue of match the @custom.navigation_layout in the `default.hbs` file.

### DIFF
--- a/default.hbs
+++ b/default.hbs
@@ -46,7 +46,7 @@
 
             <div class="gh-head-actions">
                 {{#unless @site.members_enabled}}
-                    {{^match @custom.navigation_layout "Stacked"}}
+                    {{#match @custom.navigation_layout "Stacked"}}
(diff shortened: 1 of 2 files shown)


minimaluminium:
Hi @officialrajdeepsingh, `^` instead of `#` is used for negating the helpers, and was used in these places intentionally. For example, it's telling that show a search icon when the navigation layout is _not_ "Stacked", or show a lock icon when a post is _not_ public.

officialrajdeepsingh:
I can't understand; please provide docs related to `^` on how we can use it with ghost theme development.

minimaluminium:
https://ghost.org/docs/themes/helpers/is/

For example, you can find how `^` is used with `is` helper on the bottom of the page.

officialrajdeepsingh:
Now I understand. Thank

PR #933, opened 2023-04-04
Fix problem showing dropdown after window resize

verement:
When the browser window is resized, a new click event listener is added to the window but the old one is not removed. The old one ends up preventing the newly-made dropdown from opening.

Fix this by remembering the click event listener function added to the window, and remove it before making the new dropdown after a window resize.

### DIFF
--- a/assets/js/dropdown.js
+++ b/assets/js/dropdown.js
@@ -16,6 +16,7 @@
         });
     }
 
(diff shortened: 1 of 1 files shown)


minimaluminium:
This is amazing, thanks for the contribution @verement 🙂

PR #934, opened 2023-04-05
Fix .gh-head-open accent colour in dark mode for mobile devices

kasuncfdo:
### Before
image
### After
image

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -2286,7 +2286,7 @@ html.dark-mode .footer-cta-title {
 }
 
 @media (max-width: 767px) {
-    html.dark-mode .gh-head-open #gh-head .gh-head-actions {
+    html.dark-mode #gh-head .gh-head-actions {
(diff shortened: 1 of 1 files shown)


minimaluminium:
Hey @kasuncfdo, a really nice find! Instead of leaving the CSS rules for this, we actually should've removed this bit of CSS as the action is using the accent color anyway. I removed them directly in https://github.com/TryGhost/Casper/commit/b56bec2cb8bf1e968367b2ed290587777cac8661. Thanks for your contribution!

kasuncfdo:
Thanks for the fix @minimaluminium I have another favour just found another one on the ghost admin panel's under the invite staff members section, the hover box contents are white in dark mode still not fixed in the latest version. the improvement should focus on dark mode and light mode. can you report it to the ghost 

image

minimaluminium:
Yes, of course! I've just added this to the board me manage internally. Thanks again 🙏

PR #940, opened 2023-04-28
Add prefers-reduced-motion support

Mysterious-Dev:
All is in the title ^^.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion


Mysterious-Dev:
Should I leave this PR open or is it better to close it (not useful)?

jessehouwing:
@Mysterious-Dev I merged it into my customized version of Casper, so I at least am grateful for your contribution ;).

PR #945, opened 2023-07-11
Removed spacing between full-width cards

sanne-san:
Refs https://github.com/TryGhost/Product/issues/3589


assets/css/screen.css
```diff
@@ -1347,6 +1347,11 @@ headings, text, images and lists. We deal with cards lower down. */
     margin-bottom: 0;
 }
 
+/* Remove space between full-width cards */
+.gh-content > .kg-width-full + .kg-width-full:not(.kg-width-full.kg-card-hascaption + .kg-width-full) {
```
minimaluminium, line 1351: We might want to limit it to `.page-template` here if we only want it on page templates without making it affect existing posts.

Issue #952, opened 2023-08-12
images added to posts in markdown are displayed differently than images added using the '+' button 


asenseofpradhyu:
Issue Summary
Problem:
Images added to posts in Markdown are left aligned instead of center aligned, and they do not display the caption underneath. This behavior differs from images added to posts using the '+' button in the editor.

Expected Behavior:
Images added to posts using Markdown are displayed the same as images added to posts using the '+' button in the editor. This means they should be center aligned and the caption should be displayed if a caption is provided.

Steps to Reproduce
Add a Markdown element to a post.
Add an image to the Markdown element.
Add an image element to the post not using Markdown.
Compare how the two images are displayed when the post is published.

This issue is direct reference from Ghost repo:- https://github.com/TryGhost/Ghost/issues/16685

PR #953, opened 2023-08-12
[Fix] images added to posts in markdown are displayed differently than images added using the '+' button #952

asenseofpradhyu:
Fix Issue:- #952 

Problem:
Images added to posts in Markdown are left-aligned instead of center aligned, and they do not display the caption underneath. This behavior differs from images added to posts using the '+' button in the editor.

Screenshot 2023-08-12 at 5 50 04 PM

Fix:
Images added to posts using Markdown are displayed the same as images added to posts using the '+' button in the editor. This means they should be center aligned and the caption should be displayed if a caption is provided.

Screenshot 2023-08-12 at 5 50 50 PM

### DIFF
--- a/assets/css/screen.css
+++ b/assets/css/screen.css
@@ -1401,6 +1401,10 @@ is the very first element in the post content */
     line-height: 1.6em;
 }
 
+.gh-content > p img {
+  margin: 0 auto;
+}
+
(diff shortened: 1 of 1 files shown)


minimaluminium:
This is great, thanks!

PR #975, opened 2024-02-14
Update Twitter logo to use official X logo SVG

Dartegnian:
Twitter has rebranded to X, SVG from https://about.twitter.com/en/who-we-are/brand-toolkit


Dartegnian:
Here's a screenshot of what it looks like displayed alongside a complete bio.

image

PR #1002, opened 2025-08-12
fix: author needs to be replaced with primary_author for gscan to pass

jaybuidl:
Without this fix, the theme cannot be imported into Ghost v6 and `gscan` fails.

```shell
$ yarn test

Checking theme compatibility...

Your theme has 2 errors!
-----

Errors
------
Important to fix, functionality may be degraded.

- Error: Templates must contain valid Handlebars
Affected Files: author.hbs

- Error: No unknown global helper used
Affected Files: author.hbs


```

### DIFF
--- a/author.hbs
+++ b/author.hbs
@@ -6,7 +6,7 @@
 
     <div class="post-feed">
 
-        {{#author}}
+        {{#primary_author}}
         <section class="post-card post-card-large">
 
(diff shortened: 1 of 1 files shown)


JohnONolan:
can't reproduce - was able to upload without any gscan errors

